PR #12, opened 2023-06-21
Modify the build instructions

XingerTang:
Simplified the instructions related to the submodule, and added a line of code to install the ``sphinx`` before running ``
./build_pipeline.sh``.


gregorgorjanc:
@XingerTang thanks for this and for adding sphinx. Since sphinx is now provided, will the build create docs too or shall we add more instructions to build the doc?

XingerTang:
@gregorgorjanc Yes, the build will create the docs, the instructions are already included in the build.

PR #17, opened 2023-06-28
Update publish.yml

XingerTang:
I also accidentally updated the ``AlphaPeel.zip``. But I believe it won't change anything.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,6 +8,7 @@ jobs:
         permissions:
           contents: read
           id-token: write
+        environment: release
 
         steps:
             - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


AlphaPeel.zip
gregorgorjanc, line 1: @XingerTang I am wondering if once we get PyPi up and running, should we be pushing these ZIP files into the Git repository? What do other people/projects do? I will accept this now and we can remove this bundle later.

gregorgorjanc (approved):

AlphaPeel.zip
XingerTang: @gregorgorjanc I checked a few projects and indeed we won't need the zip files anymore once everything is set up.

Issue #30, opened 2023-07-05
Add .gitattributes to clarify programming language


gregorgorjanc:
<img width="689" alt="Screenshot 2023-07-05 at 15 31 08" src="https://github.com/AlphaGenes/AlphaPeel/assets/725834/c28d6294-284f-478b-99f8-912c8588ba5c">

https://github.com/github-linguist/linguist/blob/master/docs/overrides.md

XingerTang:
TODO: fix wrong filename

PR #35, opened 2023-07-06
Add .gitattribute to override language

XingerTang:
#30 
It seems that the language calculation is not available on the forked repository (it is displayed but I cannot access the details). As far as I can see, the portion of ``Roff`` is caused by the misclassification of the two output files of the tests. I add a line to ``.gitattribute`` to classify them as documentation and modified the two files to trigger the recalculation. But I am still confused that [the two files](https://github.com/search?q=repo%3AAlphaGenes%2FAlphaPeel++language%3ARoff&type=code) are too small to contribute to 97.8% of the repository.


gregorgorjanc:
@XingerTang I know what you mean - I looked into this yesterday and gave up and opened an issue;)

README.md
```diff
@@ -24,10 +24,6 @@ AlphaPeel is available on [PyPI](https://pypi.org/project/AlphaPeel):
 
     pip install AlphaPeel
 
-## User guide
```
gregorgorjanc, line 27: @XingerTang was this omission/deletion by accident? I do think we want to point directly to docs - I am aware that PyPI site will be showing that too, but best to give direct link in the README too.

gregorgorjanc (changes requested):

README.md
XingerTang: @gregorgorjanc That is intended since there is already one user guide link from line 7 to line 9.

gregorgorjanc (approved):

PR #53, opened 2023-07-20
Implement functional tests with pytest

XingerTang:
Implement ``pytest`` as discussed in AlphaGenes/AlphaPlantImpute2#4.

I tried to implement the first two functional tests of ``tests/functional_tests/runTests.sh`` with ``pytest`` and added it to the ``tests`` workflow, so it will be triggered whenever there is a push or a pull request. 

The reason I stopped at the second test is that, from the third test, ``runTests.sh`` tries to use a package called ``plink`` which I'm not sure what it is and I guess it is [this one](https://www.cog-genomics.org/plink2/). 

I tried the package I linked above and it turns out that if it is the one then it should be placed in the ``tests/functional_tests/`` folder and the command calling it in ``runTests.sh`` also needs to be modified a little bit to make it executable on my laptop. But still, it keeps raising errors 

```
...
Logging to test3/outputs/output.called.0.1.log.
Options in effect:
  --bfile test3/outputs/output.called.0.1
  --out test3/outputs/output.called.0.1
  --real-ref-alleles
  --recode A

16384 MB RAM detected; reserving 8192 MB for main workspace.
Error: Failed to open test3/outputs/output.called.0.1.bed.
PLINK v1.90b7 64-bit (16 Jan 2023)             www.cog-genomics.org/plink/1.9/
(C) 2005-2023 Shaun Purcell, Christopher Chang   GNU General Public License v3
Logging to test3/outputs/output.called.0.99.log.
Options in effect:
  --bfile test3/outputs/output.called.0.99
  --out test3/outputs/output.called.0.99
  --real-ref-alleles
  --recode A

16384 MB RAM detected; reserving 8192 MB for main workspace.
Error: Failed to open test3/outputs/output.called.0.99.bed.
...
```

Apart from that, there are also other errors that are raised by running ``runTests.sh``.

```
/Users/evie/mambaforge/lib/python3.10/site-packages/tinypeel/tinyhouse/InputOutput.py:418: UserWarning: The module alphaplinkpython was not found. Plink files cannot be written out and will be ignored.
  warnings.warn("The module alphaplinkpython was not found. Plink files cannot be written out and will be ignored.")
```

What I think that might be the cause is these tests are written in the early stage of development and the developers hadn't decided how to handle binary files. We now have an option of ``-bfile`` of ``AlphaPeel`` to handle binary files, which doesn't exist in the commit that introduced the test files.

And what concerns me is that even though the ``runTests.sh`` keeps raising errors, it still outputs **Success** at the end. I think we need to write better tests to capture the bugs in the code.


XingerTang:
@gregorgorjanc 
Please see the above. 

I also implemented test 7 in the most recent commit. As test 3 requires ``plink``, test 5 has no data, and test 6 has no code, what I can do later is implement test 4 and test 8, and then figure out how to deal with test 3.

gregorgorjanc:
@XingerTang thanks for these efforts!

Yeah, the plink code part is problematic - Andrew/David told me that it only worked on Linu/Unix for them so I never looked/tested/ran any of plink functionality - you did find the right site for it - plink is a very popular tool for working with genomic data - it's very efficient and easy to use, which makes it popular - many people have data in plink format, which is why Andrew/David did some work on that, but it makes me wonder if we should remove that functionality, until we are content that it actually works!?

Does this help? Where does this additional knowledge leave you with respect to tests? Should we omit plink tests for now?

XingerTang:
@gregorgorjanc Yes, it helps a lot. However, I think I need to dig a little deeper into it (especially how ``AlphaPeel`` handle it now) to determine if we should remove it or improve it. I will start by focusing on implementing the other tests and also take some time to explore the "plink" files.

XingerTang:
@gregorgorjanc Hello! I managed to download the ``plink``, and it seems to work fine. While I encountered some issues downloading it because of my operating system, I think I can write a guide about it later. But the real problem is not the ``plink``, it is a module called [``alphaplinkpython``](https://pypi.org/project/alphaplinkpython/). 

The test 3 is using an output option called ``-binary_call_files`` which is dependent on ``alphaplinkpython``, but this dependency is not specified anywhere. Moreover, when I was trying to download ``alphaplinkpython``, I encountered several errors. One is its dependency on ``Cyphon``, which can be fixed by simply downloading the ``Cyphon``. While the other one is an error raised by running the ``setup.py`` which tries to match a missing file. It seems like there is some issue inside the package ``alphaplinkpython``, and from the PyPI page, it is developed by [David](david.wilson@roslin.ed.ac.uk). I think we may need to contact him for more advice to solve this issue.

.github/workflows/tests.yml
```diff
@@ -17,3 +17,28 @@ jobs:
             
             - uses: pre-commit/action@v3.0.0
               name: Run pre-commit
+
+            - name: Install pypa/build
```
gregorgorjanc, line 21: `pypa` or `pypi`?

tests/functional_tests/test1/outputs/output.called.0.1
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test1/outputs/output.dosages
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test1/outputs/output.genoError
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test1/outputs/output.maf
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test1/outputs/output.seg
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test1/outputs/output.seqError
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test2/outputs/output.called.0.1
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test2/outputs/output.dosages
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test2/outputs/output.genoError
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test2/outputs/output.maf
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test2/outputs/output.seg
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test3c/outputs/output.called.0.9.log
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

tests/functional_tests/test3c/outputs/output.called.0.9.fam
gregorgorjanc, line 1: Do we want these files in the repo? I am not saying not, just asking!

Issue #54, opened 2023-08-22
Estmaf impact on accuracy across generations


XingerTang:
The accuracy tests show that the accuracy is the lowest with the ``estmaf`` option,

```
tests/accuracy_tests/run_accu_test.py ---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_multi
Accuracy:  0.98
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_estsrrors_estmaf_multi
Accuracy:  0.978
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_seq_multi
Accuracy:  0.907
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_estmaf_multi
Accuracy:  0.863
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_hybrid_single
Accuracy:  0.979
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_hybrid_seq_single
Accuracy:  0.914
.                              
```
To investigate more about this, we want more information about the accuracy of each generation.

XingerTang:
With the fixed test name of estmaf single test, we have the accuracies among all the tests:
```
tests/accuracy_tests/run_accu_test.py ---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_multi
Accuracy:  0.98
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_estsrrors_estmaf_multi
Accuracy:  0.978
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_seq_multi
Accuracy:  0.907
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_estmaf_multi
Accuracy:  0.977
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_estmaf_single
Accuracy:  0.863
.---------------------------------------------------------------------------------------------------------
nodeid: tests/accuracy_tests/run_accu_test.py::test_hybrid_single
[...]

PR #59, opened 2023-08-25
Add accuracy tests & cross-platform tests workflow

XingerTang:
#54


.github/workflows/tests.yml
```diff
@@ -33,12 +36,13 @@ jobs:
               run: python3 -m build
 
             - name: Install AlphaPeel
-              run: pip install dist/*.whl
+              run: pip install dist/AlphaPeel-1.1.3-py3-none-any.whl
```
gregorgorjanc, line 39: Can we make this version agnostic?

gregorgorjanc:
Minor ask above @XingerTang ;)

.github/workflows/tests.yml
XingerTang: @gregorgorjanc The wildcard works fine on MacOS and Ubuntu but not on Windows. I couldn't figure out why so I just hardcoded the file name.

gregorgorjanc: Let’s worry about this later then

Issue #58, opened 2023-08-22
Add/fix test6 for sex-chrom functionality


gregorgorjanc:

XingerTang:
The problem with test 6 is even though it has both test code and data, it does not specify how to check it. Normally the tests should provide a file contains a subset of the expected output, so maybe it is also the case for the test 6.

output file to check:
```
genotypes_0 0 0 0 0 0
genotypes_2 2 2 2 2 2
genotypes_F 1 1 1 1 1
genotypes_M 2 2 2 2 2
sequence_0 0 0 0 0 0
sequence_2 2 2 2 2 2
sequence_F0 1 1 1 1 1
sequence_F9 1 1 1 1 1
sequence_M0 2 2 2 2 2
sequence_M9 2 2 2 2 2
```

trueGenotypes.txt
```
genotypes_0 0 0 0 0 0
genotypes_2 2 2 2 2 2
genotypes_F 1 1 1 1 1
genotypes_M 2 2 2 2 2
```

gregorgorjanc:
@XingerTang can you develop the missing file? Let me know if you struggle and we can figure this one together.

XingerTang:
@gregorgorjanc Sure!

Issue #57, opened 2023-08-22
Add/fix test5 for error rate functionality


gregorgorjanc:

gregorgorjanc:
@XingerTang let's discuss this PR https://github.com/AlphaGenes/AlphaPeel/pull/63 point here

> @gregorgorjanc For test 5, if we add some errors to the input data and compare the output to the true state, I'm wondering if the corrupted data can be recovered 100% to the true state, if not then how can we know which part of the loss of the accuracy is not error rate related and which part is caused by the input error rate. If we cannot separate these two cases then we won't know whether the effect of the input error rate is as expected. And for those that can return to the true state 100%, we can't tell if the accuracy is just a coincidence or a result of the correct error rate.
So, I prefer to prepare simple input data and compute the expected outcome by hand to see whether the computed and expected outcomes match.

Yes, this is why I highlighted this point. Let's do the following.

1) Report here the no error input and output for the example 1

2) Add some error to input and let's evaluate how the output changes when you assume that error rate is 0. This will give us wrong behaviour, but important to understand how it behaves in that edge case. I wonder how AlphaPeel behaves in such a case - see the end of this message.

3a) Take 2 and assume a tiny error rate and lets evaluate the output - is the output correct? Can we evaluate the exact result so we can test accuracy of our implementation against the true exact result? This will be hardish;)

3b) As 3a but larger error rate.

It would be useful to think about what we want to achieve here!? We want to test that genotype error rate functionality works well, so we should create a setting where data enables overwriting the observed genotype error. I have not looked sufficiently into the data example in test1. One way to make data overwrite the wrong observed genotype is to say have a father with genotype 0 (hence alleles 0 and 0) and SIZEABLE NUMBER of progeny with genotype 2 (hence alleles 1 and 1) so that progeny information overrules father's information.

How does this sound?

XingerTang:
@gregorgorjanc I think the procedure is very clear and can be used to test the error rates, however, I'm a bit concerned about how to write the test. For 2, 3a, 3b, the most we can get is the accuracy of the output result, so to test whether the AlphaPeel is working as expected, the most I can do is by asserting an inequality between the accuracies (i.e. accuracy of 3b > accuracy of 3a > accuracy of 2). I'm concerned about the possibility that the error rate functionality is working well, but the input data is so problematic that the peeling process cannot work in the expected way. However, I can try to produce such a test first to see whether it would work.

PR #63, opened 2023-08-28
Polish code with flake8 output & add functional test 5 and 6

XingerTang:
Polish code according to AlphaGenes/tinyhouse#5
Add functional tests 5 and 6 #57 #58

I fix the test 6 by adding a result-checking part.
For test 5, originally test 5 did not have any data, so I copied the data of test 1 to test 5, and I set the expected data to check against the output of the command of the input data. Thus, I don't know if the code behaved as expected, maybe we need to assess the validity of the data later.

### DIFF
--- a/src/tinypeel/Peeling/PeelingUpdates.py
+++ b/src/tinypeel/Peeling/PeelingUpdates.py
@@ -249,50 +249,50 @@ def updateSeqError_ind(counts, errors, refReads, altReads, genoProbs):
 #     setupTransmission(distance, peelingInfo)
 
 
-# def estDistanceFromBreaks(loc1, loc2, nBreaks, peelingInfo):
-#     breakPoints = np.floor(np.linspace(loc1, loc2, nBreaks)).astype(dtype=np.int64)
-#     distances = [
-#         getDistance(breakPoints[i], breakPoints[i + 1], peelingInfo)
-#         for i in range(nBreaks - 1)
-#     ]
-#     print(nBreaks, ":", distances)
-#     return sum(distances)
+def estDistanceFromBreaks(loc1, loc2, nBreaks, peelingInfo):
+    breakPoints = np.floor(np.linspace(loc1, loc2, nBreaks)).astype(dtype=np.int64)
+    distances = [
+        getDistance(breakPoints[i], breakPoints[i + 1], peelingInfo)
+        for i in range(nBreaks - 1)
+    ]
+    print(nBreaks, ":", distances)
+    return sum(distances)
 
 
-# def getDistance(loc1, loc2, peelingInfo):
-#     patSeg1 = getSumSeg(loc1, peelingInfo)
-#     patSeg2 = getSumSeg(loc2, peelingInfo)
+def getDistance(loc1, loc2, peelingInfo):
+    patSeg1 = getSumSeg(loc1, peelingInfo)
+    patSeg2 = getSumSeg(loc2, peelingInfo)
 
-#     patValid = ((patSeg1 > 0.99) | (patSeg1 < 0.01)) & (
-#         (patSeg2 > 0.99) | (patSeg2 < 0.01)
-#     )
+    # patValid = ((patSeg1 > 0.99) | (patSeg1 < 0.01)) & (
+    #     (patSeg2 > 0.99) | (patSeg2 < 0.01)
+    # )
 
-#     # matValid = ((matSeg1 > .99) | (matSeg1 < .01)) & ((matSeg2 > .99) | (matSeg2 < .01))
+    # matValid = ((matSeg1 > .99) | (matSeg1 < .01)) & ((matSeg2 > .99) | (matSeg2 < .01))
 
-#     # patRecomb = np.mean(np.abs(patSeg1[patValid] - patSeg2[patValid]))
-#     # matRecomb = np.mean(np.abs(matSeg1[matValid] - matSeg2[matValid]))
+    # patRecomb = np.mean(np.abs(patSeg1[patValid] - patSeg2[patValid]))
+    # matRecomb = np.mean(np.abs(matSeg1[matValid] - matSeg2[matValid]))
 
-#     # recomb = (patRecomb + matRecomb)/2
+    # recomb = (patRecomb + matRecomb)/2
 
-#     # difference = np.abs(np.round(patSeg1[patValid]) - np.round(patSeg2[patValid]))
-#     # return np.mean(difference)
+    # difference = np.abs(np.round(patSeg1[patValid]) - np.round(patSeg2[patValid]))
+    # return np.mean(difference)
 
-#     entropy1 = 1 - (-patSeg1 * np.log2(patSeg1) - (1 - patSeg1) * np.log2(1 - patSeg1))
-#     entropy2 = 1 - (-patSeg2 * np.log2(patSeg2) - (1 - patSeg2) * np.log2(1 - patSeg2))
+    entropy1 = 1 - (-patSeg1 * np.log2(patSeg1) - (1 - patSeg1) * np.log2(1 - patSeg1))
+    entropy2 = 1 - (-patSeg2 * np.log2(patSeg2) - (1 - patSeg2) * np.log2(1 - patSeg2))
 
-#     difference = patSeg1 * (1 - patSeg2) + (1 - patSeg1) * patSeg2
-#     est = np.sum(entropy1 * entropy2 * difference) / np.sum(entropy1 * entropy2)
-#     # return est
-#     return haldane(est)
-
-
-# def getSumSeg(loc, peelingInfo):
-#     seg = peelingInfo.segregation[:, :, loc]
-#     sumSeg = np.sum(seg, 1)
-#     patSeg = (seg[:, 2] + seg[:, 3]) / sumSeg
-#     # matSeg = (seg[:,1] + seg[:,3])/sumSeg
-#     return patSeg
+    difference = patSeg1 * (1 - patSeg2) + (1 - patSeg1) * patSeg2
+    est = np.sum(entropy1 * entropy2 * difference) / np.sum(entropy1 * entropy2)
+    # return est
+    return haldane(est)
 
 
-# def haldane(difference):
-#     return -np.log(1.0 - 2.0 * difference) / 2.0
+def getSumSeg(loc, peelingInfo):
+    seg = peelingInfo.segregation[:, :, loc]
+    sumSeg = np.sum(seg, 1)
+    patSeg = (seg[:, 2] + seg[:, 3]) / sumSeg
+    # matSeg = (seg[:,1] + seg[:,3])/sumSeg
+    return patSeg
+
+
+def haldane(difference):
+    return -np.log(1.0 - 2.0 * difference) / 2.0


src/tinypeel/Peeling/PeelingUpdates.py
```diff
@@ -241,64 +235,64 @@ def updateSeqError_ind(counts, errors, refReads, altReads, genoProbs):
...
+#     seg = peelingInfo.segregation[:, :, loc]
+#     sumSeg = np.sum(seg, 1)
+#     patSeg = (seg[:, 2] + seg[:, 3]) / sumSeg
+#     # matSeg = (seg[:,1] + seg[:,3])/sumSeg
+#     return patSeg
 
 
-def haldane(difference):
-    return -np.log(1.0 - 2.0 * difference) / 2.0
+# def haldane(difference):
```
gregorgorjanc, line 297: @XingerTang Is there a reason why you are commenting out the `haldane` function?

src/tinypeel/Peeling/PeelingUpdates.py
```diff
@@ -241,64 +235,64 @@ def updateSeqError_ind(counts, errors, refReads, altReads, genoProbs):
...
+#     return haldane(est)
 
 
-def getSumSeg(loc, peelingInfo):
-    seg = peelingInfo.segregation[:, :, loc]
-    sumSeg = np.sum(seg, 1)
-    patSeg = (seg[:, 2] + seg[:, 3]) / sumSeg
-    # matSeg = (seg[:,1] + seg[:,3])/sumSeg
-    return patSeg
+# def getSumSeg(loc, peelingInfo):
```
gregorgorjanc, line 289: @XingerTang Is there a reason why you are commenting out the `getSumSeg` function? Because it is not used?

src/tinypeel/Peeling/PeelingUpdates.py
```diff
@@ -241,64 +235,64 @@ def updateSeqError_ind(counts, errors, refReads, altReads, genoProbs):
 #
 
 
-def updateSeg(peelingInfo):
```
gregorgorjanc, line 244: @XingerTang these edits to `getNewton()` and related functions in this `PeelingUpdates.py` file should not be in this PR - best to keep each PR simple and doing mostly one thing - polishing flake8 and tests is fine, but you are doing something else here,

gregorgorjanc:
> I fix the test 6 by adding a result-checking part.

Test 6 is for the sex-chromosome functionality (https://github.com/AlphaGenes/AlphaPeel/issues/58). Do we have sex-chromosome genotype or sequence data - I don't see these files in this PR (are they already in the git repo from before)?

> For test 5, originally test 5 did not have any data, so I copied the data of test 1 to test 5, and I set the expected data to check against the output of the command of the input data. Thus, I don't know if the code behaved as expected, maybe we need to assess the validity of the data later.

Yeah, for test 5 we should look at the error-rate functionality (https://github.com/AlphaGenes/AlphaPeel/issues/57), so we need to discuss how to do a reasonable testing case! Do you have any suggestion how to do this in a simple way? Perhaps adding some errors into the input data compared to the true data and test that AlphaPeel recovers the true state?

src/tinypeel/Peeling/PeelingUpdates.py
XingerTang: @gregorgorjanc I comment this part because this whole part of the code is broken (more details about it in the report I wrote), and `updateSeg()` is trying to call a function `setupTransmission()` which is not defined so the flake8 complained about this. So, it is flake8 related but I did do some extra here, I can uncomment the functions other than `setupTransmission()` if it is better.

XingerTang:
> > I fix the test 6 by adding a result-checking part.
> 
> Test 6 is for the sex-chromosome functionality (#58). Do we have sex-chromosome genotype or sequence data - I don't see these files in this PR (are they already in the git repo from before)?
> 
> > For test 5, originally test 5 did not have any data, so I copied the data of test 1 to test 5, and I set the expected data to check against the output of the command of the input data. Thus, I don't know if the code behaved as expected, maybe we need to assess the validity of the data later.
> 
> Yeah, for test 5 we should look at the error-rate functionality (#57), so we need to discuss how to do a reasonable testing case! Do you have any suggestion how to do this in a simple way? Perhaps adding some errors into the input data compared to the true data and test that AlphaPeel recovers the true state?

@gregorgorjanc Fort test 6, the sex-related part is from the command and the input file `pedigree.txt`:
```
AlphaPeel -genotypes test6/genotypes.txt \
                          -seqfile test6/seqfile.txt \
                          -pedigree test6/pedigree.txt \
                          -runType multi \
                          -calling_threshold .1 \
                          -sexchrom \
                          -out test6/outputs/output
```
where `-sexchrom` indicates that this is a sex chromosome and
```
genotypes_0 0 0 M
genotypes_2 0 0 F
genotypes_M genotypes_0 genotypes_2 M 
genotypes_F genotypes_0 genotypes_2 F
sequence_0 0 0 M
sequence_2 0 0 F
sequence_M0 sequence_0 sequence_2 M 
sequence_F0 sequence_0 sequence_2 F
sequence_M9 sequence_0 sequence_2 M 
sequence_F9 sequence_0 sequence_2 F
```
gives the sex information.

XingerTang:
@gregorgorjanc For test 5, if we add some errors to the input data and compare the output to the true state, I'm wondering if the corrupted data can be recovered 100% to the true state, if not then how can we know which part of the loss of the accuracy is not error rate related and which part is caused by the input error rate. If we cannot separate these two cases then we won't know whether the effect of the input error rate is as expected. And for those that can return to the true state 100%, we can't tell if the accuracy is just a coincidence or a result of the correct error rate. 
So, I prefer to prepare simple input data and compute the expected outcome by hand to see whether the computed and expected outcomes match.

src/tinypeel/Peeling/PeelingUpdates.py
gregorgorjanc: > @gregorgorjanc I comment this part because this whole part of the code is broken (more details about it in the report I wrote), and `updateSeg()` is trying to call a function `setupTransmission()` which is not defined so the flake8 complained about this. So, it is flake8 related but I did do some extra here, I can uncomment the functions other than `setupTransmission()` if it is better.

I see! Hmm, odd, BUT good find!

Please open an issue so we have a history of such changes in the repository with a clear statement of what we should do and then PRs solving the issue are linked to the issues through "hash-links".

Then let's solve this as part of that other issue. Best to work one step at a time;)

gregorgorjanc:
> Fort test 6, the sex-related part is from the command and the input file `pedigree.txt`:
> 
> ```
> AlphaPeel -genotypes test6/genotypes.txt \
>                           -seqfile test6/seqfile.txt \
>                           -pedigree test6/pedigree.txt \
>                           -runType multi \
>                           -calling_threshold .1 \
>                           -sexchrom \
>                           -out test6/outputs/output
> ```
> 
> where `-sexchrom` indicates that this is a sex chromosome and
> 
> ```
> genotypes_0 0 0 M
> genotypes_2 0 0 F
> genotypes_M genotypes_0 genotypes_2 M 
> genotypes_F genotypes_0 genotypes_2 F
> sequence_0 0 0 M
> sequence_2 0 0 F
> sequence_M0 sequence_0 sequence_2 M 
> sequence_F0 sequence_0 sequence_2 F
> sequence_M9 sequence_0 sequence_2 M 
> sequence_F9 sequence_0 sequence_2 F
> ```
> 
> gives the sex information.

@XingerTang I understand, BUT have you looked if the input and output make sense in the context of sex chromosome - females have two copies of X chromosome so they can have allele dosage of 0, 1, or 2, while males have one copy of X, so they can have allele dosage of 0 or 1. At the moment, we ignore the PAR (pseudo-autosomal region) of the Y chromosome, where males can have allele dosage of 0, 1, or 2 (will open an issue for this for future work - not a priority now).

gregorgorjanc:
> @gregorgorjanc For test 5, if we add some errors to the input data and compare the output to the true state, I'm wondering if the corrupted data can be recovered 100% to the true state, if not then how can we know which part of the loss of the accuracy is not error rate related and which part is caused by the input error rate. If we cannot separate these two cases then we won't know whether the effect of the input error rate is as expected. And for those that can return to the true state 100%, we can't tell if the accuracy is just a coincidence or a result of the correct error rate.
> So, I prefer to prepare simple input data and compute the expected outcome by hand to see whether the computed and expected outcomes match.

I moved this discussion to https://github.com/AlphaGenes/AlphaPeel/issues/57

XingerTang:
> > Fort test 6, the sex-related part is from the command and the input file `pedigree.txt`:
> > ```
> > AlphaPeel -genotypes test6/genotypes.txt \
> >                           -seqfile test6/seqfile.txt \
> >                           -pedigree test6/pedigree.txt \
> >                           -runType multi \
> >                           -calling_threshold .1 \
> >                           -sexchrom \
> >                           -out test6/outputs/output
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where `-sexchrom` indicates that this is a sex chromosome and
> > ```
> > genotypes_0 0 0 M
> > genotypes_2 0 0 F
> > genotypes_M genotypes_0 genotypes_2 M 
> > genotypes_F genotypes_0 genotypes_2 F
> > sequence_0 0 0 M
> > sequence_2 0 0 F
> > sequence_M0 sequence_0 sequence_2 M 
> > sequence_F0 sequence_0 sequence_2 F
> > sequence_M9 sequence_0 sequence_2 M 
> > sequence_F9 sequence_0 sequence_2 F
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > gives the sex information.
> 
> @XingerTang I understand, BUT have you looked if the input and output make sense in the context of sex chromosome - females have two copies of X chromosome so they can have allele dosage of 0, 1, or 2, while males have one copy of X, so they can have allele dosage of 0 or 1. At the moment, we ignore the PAR (pseudo-autosomal region) of the Y chromosome, where males can have allele dosage of 0, 1, or 2 (will open an issue for this for future work - not a priority now).

I checked the input genotypes and the sequence file, and almost every entry is filled with `0` or `9` except the `genotypes_2`, which has `2` on every locus, but the individual is a female, so the input data is fine. However, the output data (both the output generated by `AlphaPeel` and the expected output written a long time ago) contains male individuals with an allele dosage of `2`. The `AlphaPeel` documentation marked the `-sexchrom` as an experimental option, so it might be still under development and requires further refinement.

gregorgorjanc:
> I checked the input genotypes and the sequence file, and almost every entry is filled with `0` or `9` except the `genotypes_2`, which has `2` on every locus, but the individual is a female, so the input data is fine. However, the output data (both the output generated by `AlphaPeel` and the expected output written a long time ago) contains male individuals with an allele dosage of `2`. The `AlphaPeel` documentation marked the `-sexchrom` as an experimental option, so it might be still under development and requires further refinement.

@XingerTang ok, I propose to get this PR in and we open a new issue to address the sexchrom output, but I think this also raises the fact that we don't have sufficient information about these tests to really trust them.

gregorgorjanc:
@XingerTang resolve the stuff mentioned in https://github.com/AlphaGenes/AlphaPeel/pull/63#discussion_r1308308145 and I will merge this PR.

PR #86, opened 2023-09-29
Add new test files for the accuracy tests and Update tests for the new option names

XingerTang:
Added the simulated test data for the accuracy tests, but before implementing the updated test data, there are a few things left to decide:

- For the hybrid peeling, are we going to continue with the same subsets of the sites as before (200 evenly spaced loci for every individual), or will we use another subset of sites
- The way to check the accuracy of the segregation file

PR #119, opened 2023-10-17
Changing more argument names

AprilYUZhang:


XingerTang (changes requested):
@AprilYUZhang ``calling_threshold`` cannot be directly replaced by the ``geno_threshold``, if you look at the code for it, you will find that it is for both the haplotypes and genotypes output, so you would need to split the functionality of it to ``hap_threshold`` and ``geno_threshold``.

I've done it before in [this commit](https://github.com/AlphaGenes/AlphaPeel/pull/86/commits/6d0f9863e76bb4c789a7f75d878d5cc365e1f7ba) but I didn't have a chance to do a PR for it. I think it could be a reference for you to modify the code.

gregorgorjanc:
@XingerTang can you please work with @AprilYUZhang on revising these changes? Note that we might need to change `tinyhouse` too!

XingerTang:
> @XingerTang can you please work with @AprilYUZhang on revising these changes? Note that we might need to change `tinyhouse` too!

@gregorgorjanc Sure! But still, I think the modification in the `tinyhouse` is unnecessary as we can redefine the functions/variables in the AlphaPeel then we can just modify things inside AlphaPeel.

gregorgorjanc:
> I think the modification in the `tinyhouse` is unnecessary as we can redefine the functions/variables in the AlphaPeel then we can just modify things inside AlphaPeel.

Good point That's probably safer! We do need to make sure we are consistent since some documentation/help sists in `tinypeel` if I am not mistaken.

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -76,25 +76,25 @@ def writeGenotypes(pedigree, genoProbFunc):
...
-    if args.calling_threshold is not None:
-        for thresh in args.calling_threshold:
+    if args.geno_threshold is not None:
+        for thresh in args.geno_threshold:
             if args.binary_call_files:
                 writeBinaryCalledGenotypes(
                     pedigree, genoProbFunc, args.out + ".called." + str(thresh), thresh
                 )
-            if not args.binary_call_files:
+            if args.geno:
```
XingerTang, line 88: Please modify line 88 to `if not args.binary_call_files and args.geno:`

docs/source/usage.rst
```diff
@@ -62,15 +62,17 @@ Output Arguments
       -hap                  Flag to call and write out the haplotypes.
       -geno                 Flag to call and write out the genotypes.
       -geno_threshold [GENO_THRESHOLD [GENO_THRESHOLD ...]]
-                            Genotype calling threshold(s). Multiple space
-                            separated values allowed. Use .3 for best guess
-                            genotype.
+                            Genotype calling threshold(s). Multiple space separated values allowed.
+                            Any value less than 1 / 3 would be replaced by 1 / 3.
```
gregorgorjanc, line 67: @AprilYUZhang: Change `Any value less than 1 / 3 would be replaced by 1 / 3.` to `Value less than 1 / 3 will be replaced by 1 / 3.`

docs/source/usage.rst
```diff
@@ -62,15 +62,17 @@ Output Arguments
...
+      -hap_threshold [HAPS_CALL_THRESHOLD [HAPS_CALL_THRESHOLD]...]
+                            Haplotype calling threshold(s). Multiple space separated values allowed.
+                            Any value less than 0.5 would be replaced by 0.5.
```
gregorgorjanc, line 70: @AprilYUZhang: Change `Any value less than 0.5 would be replaced by 0.5.` to `Value less than 1/2 will be replaced by 1/2.`

src/tinypeel/tinypeel.py
```diff
@@ -282,7 +282,17 @@ def getArgs():
         required=False,
         type=float,
         nargs="*",
-        help="Genotype calling threshold(s). Multiple space separated values allowed. Use. .3 for best guess genotype.",
+        help="Genotype calling threshold(s). Multiple space separated values allowed. \
+         Any value less than 1 / 3 would be replaced by 1 / 3.",
```
gregorgorjanc, line 291: @AprilYUZhang sync with the above suggestion

src/tinypeel/tinypeel.py
```diff
@@ -282,7 +282,17 @@ def getArgs():
...
+    )
+    output_parser.add_argument(
+        "-hap_threshold",
+        default=None,
+        required=False,
+        type=float,
+        nargs="*",
+        help="Haplotype calling threshold(s). Multiple space separated values allowed. \
+          Any value less than 0.5 would be replaced by 0.5.",
```
gregorgorjanc, line 300: @AprilYUZhang sync with the above suggestion

docs/source/usage.rst
```diff
@@ -70,7 +71,7 @@ Output Arguments
       -binary_call_files    Flag to write out the called genotype files as a
                             binary plink output [Not yet implemented].
 
-By default |Software| produces a dosage file, a segregation files and two model parameter files (genotyping error and recombination rate). Creation of these files can be suppressed with the ``-no_dosage``, and ``-no_params`` options. |Software| can also write out the genotype probability file (.phased_geno_prob.txt) with the `-phased_geno_prob` argument and the segregation probability file (.seg_prob.txt) with the `-seg_prob` argument.
+By default |Software| produces a dosage file, a segregation files and two model parameter files (genotyping error and recombination rate). Creation of these files can be suppressed with the ``-no_dosage``, and ``-no_params`` options. |Software| can also write out the phased genotype probability file (.phased_geno_prob.txt) with the `-phased_geno_prob` argument and the segregation probability file (.seg_prob.txt) with the `-seg_prob` argument.
```
gregorgorjanc, line 74: @AprilYUZhang we should change `By default |Software| produces a dosage file, a segregation files and two model parameter files (genotyping error and recombination rate). ` to `By default |Software| produces a dosage file and two model parameter files (genotype error rate and recombination rate).` But this will likely be addressed in the next commits.

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -118,6 +119,30 @@ def writeGenoProbs(pedigree, genoProbFunc, outputFile):
                 )
 
 
+def writeGenoProbs(pedigree, genoProbFunc, outputFile):
+    with open(outputFile, "w+") as f:
+        for idx, ind in pedigree.writeOrder():
+            matrix = genoProbFunc(ind.idn)
+            f.write("\n")
+            for i in range(matrix.shape[0]):
```
gregorgorjanc, line 127: @AprilYUZhang My Python is at an entry-level, but I think this loop is wrong. We will loop from 0 to 4 and we effectively want to sum up rows 0&1 and 2&3. At the moment you are summing up 1+2 (instead of 0+1) and you are only outputting 2 in the elif part (instead of 2+3). Does this make sense?

XingerTang: @gregorgorjanc What you interpret from the code is absolutely correct, but for the matrix of genotype probability, each individual has 4 lines to represent the probabilities of having aa, aA, Aa, AA alleles respectively. So, if we want the unphased genotype probability, we need to add up the second and the third row for each individual.

gregorgorjanc: @XingerTang yeah, you are right, so we should output 0, 1+2, and 3, which is what the code does! Doh. Sorry for the noise @AprilYUZhang ;)

To avoid such blunders, it would help to add some comments to the code. What about this:

```
def writeGenoProbs(pedigree, genoProbFunc, outputFile):
     with open(outputFile, "w+") as f:
         for idx, ind in pedigree.writeOrder():
             matrix = genoProbFunc(ind.idn)
             f.write("\n")
             for i in range(matrix.shape[0]):
                if i == 1: # Add up probabilities for aA and Aa
                     f.write(
                         ind.idx
                         + " "
                         + " ".join(
                             map("{:.4f}".format, matrix[i, :] + matrix[i + 1, :])
                         )
                         + "\n"
                     )
                 elif i != 2: # Print probabilities for aa and AA
                     f.write(
                         ind.idx
                         + " "
                         + " ".join(map("{:.4f}".format, matrix[i, :]))
                         + "\n"
                     )
```

XingerTang: @gregorgorjanc I think the comments are very clear.

src/tinypeel/Peeling/PeelingIO.py
AprilYUZhang: Update.

docs/source/usage.rst
AprilYUZhang: have updated.

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -76,30 +76,39 @@ def writeGenotypes(pedigree, genoProbFunc):
...
+    if args.geno_threshold is not None:
+        for thresh in args.geno_threshold:
+            if thresh < 1 / 3:
+                thresh = 1 / 3
             if args.binary_call_files:
                 writeBinaryCalledGenotypes(
                     pedigree, genoProbFunc, args.out + ".called." + str(thresh), thresh
                 )
-            if not args.binary_call_files:
+            if not args.binary_call_files and args.geno:
```
gregorgorjanc, line 91: @AprilYUZhang probably best to remove `and args.geno` here (I appreciate that @XingerTang said to do it), and use it above in changing `if args.geno_threshold is not None:` to `if args.geno and args.geno_threshold is not None:`

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -76,30 +76,39 @@ def writeGenotypes(pedigree, genoProbFunc):
...
-            if args.hap:
-                writeCalledPhase(
                     pedigree,
                     genoProbFunc,
-                    args.out + ".called_phase." + str(thresh),
+                    args.out + ".geno_" + str(thresh) + ".txt",
                     thresh,
                 )
 
+            if args.hap_threshold is not None and args.hap:
```
gregorgorjanc, line 99: @AprilYUZhang probably best to change this line to  `if args.hap and args.hap_threshold is not None:` - it's easier to follow, but of course does the same thing

src/tinypeel/Peeling/PeelingIO.py
gregorgorjanc: @AprilYUZhang don't forget to add the comments to the code

gregorgorjanc (changes requested):
@AprilYUZhang see some of the requested changes in the above comments

src/tinypeel/Peeling/PeelingIO.py
XingerTang: I agree with @gregorgorjanc, but I noticed another place could also be improved at the same time. I think it might be better to change from ``if args.geno and args.geno_threshold is not None:`` to ``if args.geno and args.geno_threshold:`` . 

I wrote a small testing file for the arguments input that allows multiple float inputs. 

```testing.py
import argparse

parser = argparse.ArgumentParser(description = 'Test')
parser.add_argument('-arg1',
                    default=None,
                    required=False,
                    type=float,
                    nargs="*"
                    )
args = vars(parser.parse_args())
if args["arg1"] is not None:
    print(args["arg1"])
else:
    print("expected")
```
If we now run the following code
```
python /Users/evie/test/src/test/testing.py -arg1
```
Then it would give ``[]``, instead of notice that there is actually no input.

But if the code is as follows (I only change the if condition)
```testing.py
import argparse

parser = argparse.ArgumentParser(description = 'Test')
parser.add_argument('-arg1',
                    default=None,
                    required=False,
                    type=float,
                    nargs="*"
                    )
args = vars(parser.parse_args())
if args["arg1"]:
    print(args["arg1"])
else:
    print("expected")
```
then ``python /Users/evie/test/src/test/testing.py -arg1`` would gives ``expected``.

gregorgorjanc: @XingerTang @AprilYUZhang I will follow your advice here;)

src/tinypeel/Peeling/PeelingIO.py
AprilYUZhang: I agree; I made similar changes as `if args.geno and args.geno_threshold` .

XingerTang (approved):

XingerTang:
@gregorgorjanc I think everything is good now, could you review these changes again?

gregorgorjanc (approved):
Good job @AprilYUZhang! Thanks!

PR #121, opened 2023-11-08
Devel-stage 2

AprilYUZhang:


AprilYUZhang:
I pulled changes before I modified, but there is a strange point. Both I and @XingerTang found some commits is ahead AlhphaGenes:devel. Maybe it caused by squash, like this problem (https://github.com/orgs/community/discussions/23249). in terms that the file `est.estmaf.geno` can't be found, I guess script related to test are not updated.

XingerTang:
@gregorgorjanc 
We did some research on squashing and found that it would be better to not squash the commits for the pull requests from our `devel` branches. Here is the link for the corresponding documentation: [Squashing and merging a long-running branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squashing-and-merging-a-long-running-branch)

Overall, the squashed commit on AlphaGenes/AlphaPeel devel branch would be considered differently from the original commits on the devel branches of our repositories, so each time we create a new pull request it would take all the commits that are already squashed along with the new commits we added specifically for the new pull request. And this workflow would make merge conflicts more likely. 

So it would be better if we do the squashes on our repositories and then open pull requests for the squashed commit instead of making a squashed commit of the pull requests. 

In this case, if we want to do this properly, maybe you should reset to the commit before the squashed commits. Then we would each squash the commits locally and open some new pull requests.

XingerTang:
@gregorgorjanc The tests are failing just because the tests are not up to date with the changes of the most recent commits. I will update the tests soon.

gregorgorjanc:
> So it would be better if we do the squashes on our repositories and then open pull requests for the squashed commit instead of making a squashed commit of the pull requests.

@XingerTang yeah, this is how I have seen it done in other projects and assumed that if I squash on the upstream side you will get the same on your end when you pull. To avoid the current mess, @AprilYUZhang could you save your local repository, delete your fork, fork again, clone again, apply your recent changes, commit, push, and reopen pull request. Then before we merge in the pull request, I will ask you to squash commits into one.

AprilYUZhang:
OK, I will try

Issue #113, opened 2023-10-10
Update tests according to the changes in the options names and file names


XingerTang:
@gregorgorjanc @AudreyAAMartin @AprilYUZhang

Here is a summary of the changes in the option names and the file names 

|Old option name|Updated option name|Old file name|Updated file name|Complete|Issue|
|--|--|--|--|--|--|
|-|-|dosages|dosage.txt|√|#90|
|no_dosages|no_dosage|-|-|√|#90|
|-|seg_prob|seg|seg_prob.txt|√|#96,#98|
|no_seg|-|-|-|√ |#96,#98|
|calling_threshold|geno_threshold|-|-|√ |#82|
|-|geno|called.x|geno_x.txt|√ |#82|
|-|hap_threshold|-|-|√ |#89,#82|
|call_phase|hap|called_phase.x|hap_x.txt|√ |#89,#82|
|-|geno_prob|-|geno_prob.txt|√|#93|
|haps|phased_geno_prob|haps|phased_geno_prob.txt|√|#92|
|pedigree|ped_file|INPUT FILE|INPUT FILE|√|#109|
|genotypes|geno_file|INPUT FILE|INPUT FILE|√|#109|
|seqfile|seq_file|INPUT FILE|INPUT FILE|√|#109|
|bfile|plink_file|INPUT FILE|INPUT FILE|√|#109|
|mapfile|map_file|INPUT FILE|INPUT FILE|√|#106|
|segmapfile|seg_map_file|INPUT FILE|INPUT FILE|√|#106|
|segfile|seg_file|INPUT FILE|INPUT FILE|√|#106|
|estmaf|est_alt_allele_prob|maf|alt_allele_prob.txt|√|#97,#104|
|error|geno_error_prob|-||√|#97,#102|
|seqerror|seq_error_prob|-|-|√|#97,#102|
|esterrors|est_geno_error_prob|genoError|geno_error_prob.txt|√|#97,#102,#104|
|esterrors|est_seq_error_prob|seqError|seq_error_prob.txt|√|#97,#102,#104|
|-|est_rec_prob|-|rec_prob.txt|√|#97,#104|
|length|rec_length|-|-|√|#103|
|sexchrom|sex_chrom|-|-|√|#107|
|startsnp|start_snp|-|-|√|#108|
|stopsnp|stop_snp|-|-|√|#108|
|out|out_file|-|-|√|#110|
|writekey|out_id_order|-|-|√|#110|
[...]

Issue #108, opened 2023-10-09
Rename -startsnp to -start_snp and -stopsnp to -stop_snp


gregorgorjanc:
To make all arguments consistent we should rename

- [ ] `-startsnp` to `-start_snp` 
- [ ]  `-stopsnp` to `-stop_snp`.

gregorgorjanc:
Closed as completed.

Issue #104, opened 2023-10-09
Rename the -esterrors and -estmaf


gregorgorjanc:
Currently we have `-esterrors` and `-estmaf` to estimate genotyping error rate/prob and alternative allele rate/freq/prob. In line with https://github.com/AlphaGenes/AlphaPeel/issues/97 we should use

- [ ] `-est_geno_error_prob`
- [ ] `-est_alt_allele_prob`

possibly later also

- [ ] `-est_rec_prob`
- [ ] `-est_seq_error_prob`

gregorgorjanc:
Closing as completed.

Issue #97, opened 2023-10-09
consolidate model parameter filenames


AprilYUZhang:
- [ ] rename .maf to . alt_allele_prob
- [ ] rename .genoError to .geno_error_prob
- [ ] rename .seqError to .seq_error_prob
- [ ] write .rec_prob (at the moment it does not exist)
- [ ] add no_param

PR #122, opened 2023-11-11
Stage_2

AprilYUZhang:


docs/source/usage.rst
```diff
@@ -300,9 +302,9 @@ Example:
 Model parameter files
 =====================
 
-|Software| outputs three model parameter files, ``.maf``, ``.seqError``, ``.genoError``. These give the minor allele frequency, sequencing error rates, and genotyping error rates used. All three files contain a single column with an entry for each marker.
+|Software| outputs three model parameter files, ``.alt_allele_prob.txt``, ``.seqError``, ``.genoError``. These give the minor allele frequency, sequencing error rates, and genotyping error rates used. All three files contain a single column with an entry for each marker.
```
gregorgorjanc, line 305: @AprilYUZhang we should change docs for these error file outputs too, right?

AprilYUZhang: Yes, it will be next step. I changed the order a little.

docs/source/usage.rst
```diff
@@ -108,16 +110,18 @@ Peeling arguments
...
+The ``-geno_error_prob``, ``-seq_error_prob`` and ``-length`` arguments control some of the model parameters used in the model. ``-seq_error_prob`` must not be zero. |Software| is robust to deviations in genotyping error rate and sequencing error rate so it is not recommended to use these options unless large deviations from the default are known. Changing the ``-length`` argument to match the genetic map length can increase accuracy in some situations.
 
-The ``-esterrors`` option estimated the genotyping error rate based on observed information, this option is generally not necessary and can increase runtime. ``-estmaf`` estimates the minor allele frequency after each peeling cycle. This option can be useful if there are a large number of non-genotyped founders. 
+The ``-est_geno_error_prob`` and ``-est_seq_error_prob`` option estimated the genotyping error rate and the sequencing error rate separately based on observed information, this option is generally not necessary and can increase runtime. ``-est_alt_allele_prob`` estimates the alternative allele probability after each peeling cycle. This option can be useful if there are a large number of non-genotyped founders.
```
gregorgorjanc, line 124: @AprilYUZhang change to

```
The ``-est_geno_error_prob`` and ``-est_seq_error_prob`` options estimate the genotyping error rate and the sequencing error rate based on miss-match between observed and inferred states. This option is generally not necessary and can increase runtime. ``-est_alt_allele_prob`` estimates the alternative allele probability after each peeling cycle. This option can be useful if there are a large number of non-genotyped founders.
```

src/tinypeel/tinypeel.py
```diff
@@ -357,22 +385,28 @@ def getArgs():
...
+        action="store_true",
+        required=False,
+        help="Flag to re-estimate the sequencing error rates after each peeling cycle.",
+    )
+    peeling_control_parser.add_argument(
+        "-est_alt_allele_prob",
         action="store_true",
         required=False,
-        help="Flag to re-estimate the minor allele frequency after each peeling cycle.",
+        help="Flag to re-estimate the alternative allele probability after each peeling cycle.",
```
gregorgorjanc, line 409: @AprilYUZhang change to `probabilities` since we do it for multiple loci. Check also elsewhere.

src/tinypeel/tinypeel.py
```diff
@@ -28,9 +28,11 @@ def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
         #     print("Estimating the transmission rate is currently a disabled option")
         # PeelingUpdates.updateSeg(peelingInfo) #Option currently disabled.
 
-        if args.esterrors:
+        if args.est_geno_error_prob:
```
gregorgorjanc, line 31: @AprilYUZhang should we do an "or" statement here? `if args.est_geno... or args.est_seq...`?

docs/source/usage.rst
gregorgorjanc: @AprilYUZhang is now the time to address these file name changes (`seqErrors` and `genoErrors`)?

gregorgorjanc (changes requested):
@AprilYUZhang I requested minor tweaks

src/tinypeel/tinypeel.py
AprilYUZhang: I agree, and `if .. or ..:` is better.

AprilYUZhang:
I found a recombination rate table including human, pig, rat, and so on. Aimed at different species, we need different recombination rates, It may be useful  [Dumont BL, Payseur BA. Evolution of the genomic rate of recombination in mammals. Evolution. 62:276, 2008.](https://onlinelibrary.wiley.com/doi/10.1111/j.1558-5646.2007.00278.x)

AprilYUZhang:
In commit # 0632f47, I copied the `get_input_options()` from tinyhouse, but I just changed the input name when it parsed parameters. The reason why I didn't change the name in the `parse_dictionary`, `pedigree, genotypes, seqfile, bfile` are involved many times in the program calls of tinyhouse.

docs/source/usage.rst
```diff
@@ -108,7 +108,7 @@ Peeling arguments
                             each peeling cycle.
       -nophasefounders      A flag phase a heterozygous allele in one of the
                             founders (if such an allele can be found).
-      -sexchrom             A flag to that this is a sex chromosome. Sex needs to
+      -sex_chrom            A flag to that this is a sex chromosome. Sex needs to
```
gregorgorjanc, line 111: @AprilYUZhang please modify to `A flag to indicate that input data is for a sex chromosome. Sex needs to`

src/tinypeel/tinypeel.py
```diff
@@ -488,7 +488,7 @@ def getArgs():
         help="A flag phase a heterozygous allele in one of the founders (if such an allele can be found).",
     )
     peeling_control_parser.add_argument(
-        "-sexchrom",
+        "-sex_chrom",
         action="store_true",
         required=False,
         help="A flag to that this is a sex chromosome. Sex needs to be given in the pedigree file. This is currently an experimental option.",
```
gregorgorjanc, line 494: @AprilYUZhang use the same help string tweak as above

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -75,10 +75,10 @@ def writeOutParamaters(peelingInfo):
     np.savetxt(args.out + ".alt_allele_prob.txt", peelingInfo.alt_allele_prob, fmt="%f")
 
 
-def writeGenotypes(pedigree, genoProbFunc):
+def writeGenotypes(pedigree, genoProbFunc, isSexChrom):
```
gregorgorjanc, line 78: @AprilYUZhang I understand what you are doing here, but this is a departure from how you handled the option name changes before. Is there a particular reason that you are now picking a different strategy. Doing that makes code even more non-uniform and if you do this a couple of times, the code style becomes impossible to follow.

gregorgorjanc (changes requested):
@AprilYUZhang see a minor and a major comment

gregorgorjanc:
> I found a recombination rate table including human, pig, rat, and so on. Aimed at different species, we need different recombination rates, It may be useful [Dumont BL, Payseur BA. Evolution of the genomic rate of recombination in mammals. Evolution. 62:276, 2008.](https://onlinelibrary.wiley.com/doi/10.1111/j.1558-5646.2007.00278.x)

Thanks for sharing. There are now even newer recombination rates for some species.

gregorgorjanc:
> In commit # [0632f47](https://github.com/AlphaGenes/AlphaPeel/commit/0632f47a3492b06a9f75491f08136c41c61146ce), I copied the `get_input_options()` from tinyhouse, but I just changed the input name when it parsed parameters. The reason why I didn't change the name in the `parse_dictionary`, `pedigree, genotypes, seqfile, bfile` are involved many times in the program calls of tinyhouse.

@AprilYUZhang @XingerTang I see you are challenged quite a bit by trying to keep `tinyhouse` intact, while changing `AlphaPeel`. I spoke to Hannes today (@bryo-han) and he was of the view that you are effectively doing double work - now on AlphaPeel side and then in the future we will have to fix `tinyhouse` again. What about actually breaking `tinyhouse` for other programs? All other programs bring in `tinyhouse` dependency that is linked to a particular commit via a git submodule. So, if we push development of `tinyhouse` further, we will not immediately break other programs. Once we do the changes we need in AlphaPeel and tinyhouse we can always go an fix any other AlphaX program so that it uses the new `tinyhouse` version. This should simplify things a lot!? Thoughts?

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -68,6 +68,7 @@ def writeOutParamaters(peelingInfo):
 
     np.savetxt(args.out + ".geno_error_prob.txt", peelingInfo.genoError, fmt="%f")
     np.savetxt(args.out + ".seq_error_prob.txt", peelingInfo.seqError, fmt="%f")
+    np.savetxt(args.out + ".rec_prob.txt", np.empty((1,1)), fmt="%f") # not be realized, just as a placeholder
```
gregorgorjanc: @AprilYUZhang Please open an issue so we don't forget to come back to this code. It's easy to forget such tiny TODOs in the code. Also, we are likely storing recombination length of the chromosome and use it to calculate recombination rate between loci - it is that rate that should be outputted here. Make a note of this in the issue.

src/tinypeel/tinypeel.py
```diff
@@ -238,6 +238,36 @@ def get_probability_options():
...
+                                                                       help='A haplotype reference panel in AlphaGenes format.')
+    parse_dictionary["seqfile"] = lambda parser: parser.add_argument('-seq_file', default=None, required=False, type=str,
+                                                                     nargs="*", help='A sequence data file.')
+    parse_dictionary["pedigree"] = lambda parser: parser.add_argument('-ped_file', default=None, required=False,
+                                                                      type=str, nargs="*",
+                                                                      help='A pedigree file in AlphaGenes format.')
+    parse_dictionary["phasefile"] = lambda parser: parser.add_argument('-phasefile', default=None, required=False,
+                                                                       type=str, nargs="*",
+                                                                       help='A phase file in AlphaGenes format.')
+    parse_dictionary["startsnp"] = lambda parser: parser.add_argument('-startsnp', default=None, required=False,
```
gregorgorjanc: @AprilYUZhang in https://github.com/AlphaGenes/AlphaPeel/issues/108 we discussed to use `-start_snp` and `-stop_snp`

src/tinypeel/tinypeel.py
```diff
@@ -238,6 +238,36 @@ def get_probability_options():
...
+                                                                       help='A file in AlphaGenes format.')
+    parse_dictionary["reference"] = lambda parser: parser.add_argument('-reference', default=None, required=False,
+                                                                       type=str, nargs="*",
+                                                                       help='A haplotype reference panel in AlphaGenes format.')
+    parse_dictionary["seqfile"] = lambda parser: parser.add_argument('-seq_file', default=None, required=False, type=str,
+                                                                     nargs="*", help='A sequence data file.')
+    parse_dictionary["pedigree"] = lambda parser: parser.add_argument('-ped_file', default=None, required=False,
+                                                                      type=str, nargs="*",
+                                                                      help='A pedigree file in AlphaGenes format.')
+    parse_dictionary["phasefile"] = lambda parser: parser.add_argument('-phasefile', default=None, required=False,
```
gregorgorjanc: @AprilYUZhang I think we have missed this in our meeting - in line with the https://github.com/AlphaGenes/AlphaPeel/issues/89 we should call this `-hap_file`

src/tinypeel/tinypeel.py
```diff
@@ -238,6 +238,36 @@ def get_probability_options():
...
+                                                                       type=str, nargs="*",
+                                                                       help='A phase file in AlphaGenes format.')
```
gregorgorjanc: `A haplotype file in AlphaGenes format.`

src/tinypeel/tinypeel.py
```diff
@@ -238,6 +238,36 @@ def get_probability_options():
...
+    parse_dictionary["startsnp"] = lambda parser: parser.add_argument('-startsnp', default=None, required=False,
+                                                                      type=int,
+                                                                      help="The first marker to consider. The first marker in the file is marker '1'. Default: 1.")
+    parse_dictionary["stopsnp"] = lambda parser: parser.add_argument('-stopsnp', default=None, required=False, type=int,
```
gregorgorjanc: @AprilYUZhang as with `-start_snp`

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -68,7 +68,9 @@ def writeOutParamaters(peelingInfo):
 
     np.savetxt(args.out + ".geno_error_prob.txt", peelingInfo.genoError, fmt="%f")
     np.savetxt(args.out + ".seq_error_prob.txt", peelingInfo.seqError, fmt="%f")
-    np.savetxt(args.out + ".rec_prob.txt", np.empty((1,1)), fmt="%f") # not be realized, just as a placeholder
+    np.savetxt(
```
gregorgorjanc, line 71: @AprilYUZhang as mentioned in another comment - we should flag this line in an issue so we get back to it at some point

XingerTang:
> > In commit # [0632f47](https://github.com/AlphaGenes/AlphaPeel/commit/0632f47a3492b06a9f75491f08136c41c61146ce), I copied the `get_input_options()` from tinyhouse, but I just changed the input name when it parsed parameters. The reason why I didn't change the name in the `parse_dictionary`, `pedigree, genotypes, seqfile, bfile` are involved many times in the program calls of tinyhouse.
> 
> @AprilYUZhang @XingerTang I see you are challenged quite a bit by trying to keep `tinyhouse` intact, while changing `AlphaPeel`. I spoke to Hannes today (@bryo-han) and he was of the view that you are effectively doing double work - now on AlphaPeel side and then in the future we will have to fix `tinyhouse` again. What about actually breaking `tinyhouse` for other programs? All other programs bring in `tinyhouse` dependency that is linked to a particular commit via a git submodule. So, if we push development of `tinyhouse` further, we will not immediately break other programs. Once we do the changes we need in AlphaPeel and tinyhouse we can always go an fix any other AlphaX program so that it uses the new `tinyhouse` version. This should simplify things a lot!? Thoughts?

@gregorgorjanc 
I understand that we may need to do these changes in `tinyhouse` later as well, but @AprilYUZhang has already finished all the modifications related to the `tinyhouse`. And the update of the option names of `tinyhouse` is more difficult than that of `alphapeel` as the option names are used everywhere among the 10 files of `tinyhouse`. It would be more likely for us to miss some of them while modifying the code. 

Meanwhile, `tinyhouse` may still have some bugs unfixed, using different versions of it means we need to introduce changes to different versions of it every time we want to make an update. 

I think it would be better if we just stick with this "double work"  for a while and then when we are ready to make some big changes to the `tinyhouse`, we can do everything at once.

gregorgorjanc:
> the update of the option names of `tinyhouse` is more difficult than that of `alphapeel` as the option names are used everywhere among the 10 files of `tinyhouse`. It would be more likely for us to miss some of them while modifying the code.

@XingerTang why would this be hard? It's mostly an input and maybe some variable name change, which is less than doing extra logic, particularly in https://github.com/AlphaGenes/AlphaPeel/pull/122/commits/5821d432e533277590a2b620dfbb4d8feb76708e. The part that I am worried is that while these current changes are not large, they are adding extra logic to `AlphaPeel` and when we decide to polish `tinyhouse` we will forget that some of this extra logic should be removed from `AlphaPeel`.

XingerTang (changes requested):

XingerTang:
> > the update of the option names of `tinyhouse` is more difficult than that of `alphapeel` as the option names are used everywhere among the 10 files of `tinyhouse`. It would be more likely for us to miss some of them while modifying the code.
> 
> @XingerTang why would this be hard? It's mostly an input and maybe some variable name change, which is less than doing extra logic, particularly in [5821d43](https://github.com/AlphaGenes/AlphaPeel/commit/5821d432e533277590a2b620dfbb4d8feb76708e). The part that I am worried is that while these current changes are not large, they are adding extra logic to `AlphaPeel` and when we decide to polish `tinyhouse` we will forget that some of this extra logic should be removed from `AlphaPeel`.

It would be hard doesn't because it's hard to write anything, but the code of `AlphaPeel` and `tinyhouse` are messy, and it is hard to track where the option names go. For those option names that are adding/modifying functionality, it is not enough to do "search and replace" as we need to find everywhere it is used and understand how it is used.
In the above, @AprilYUZhang found where the errors are updated and modified the line of the condition for estimating genotype error and sequence error before that line, 
from 
```
        if args.esterrors:
            PeelingUpdates.updatePenetrance(pedigree, peelingInfo)
```
to
```
        if args.est_geno_error_prob or args.est_seq_error_prob:
            PeelingUpdates.updatePenetrance(pedigree, peelingInfo)
```
but that is not enough, we need to look up quite closely to see which line corresponds to the update of genotype error and which line corresponds to the update of sequence error inside ``PeelingUpdates.updatePenetrance(pedigree, peelingInfo)`` and add additional conditions for them, even though there is no ``args.esterrors`` in it. We need to be quite familiar with the code to do all the changes without missing anything. And, when we are facing `tinyhouse`, there are 10 files. It is easy to lose track of where and how some values are passed from one place to another, but we need to understand the uses of all of them before making changes. That is the time-consuming part. 
Also, we don't want code inside `tinyhouse` inconsistent, so we cannot modify only the `AlphaPeel` related part of `tinyhouse`. 

When we polish `tinyhouse`, we also have to check every place of `AlphaPeel` that is related to `tinyhouse` to avoid inconsistency, so we won't miss any extra logic.

gregorgorjanc:
@XingerTang @AprilYUZhang I think I understand now. Ok, let's continue and be mindful of the steps beyond the current step. I appreciate your efforts!

src/tinypeel/tinypeel.py
AprilYUZhang: Yes, these options are in the later order of the changing list.

src/tinypeel/tinypeel.py
AprilYUZhang: Ok, I will add it.

src/tinypeel/Peeling/PeelingIO.py
gregorgorjanc: @AprilYUZhang OK I propose the following - when you encounter such inconsistencies, open an issue so we can later fix them. By finding these problems, you have already spent quite a bit of time understanding the logic so if you open an issue and explain this in the issue content, then that effort is not lost and it easier to pick up at a later stage when you get back to the issue. OK?

src/tinypeel/Peeling/PeelingUpdates.py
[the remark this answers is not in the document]
AprilYUZhang: `if args.est_geno_error_prob or args.est_seq_error_prob:` doesn't work, right? I'm not sure if you mean  problem happens to  `def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
 `.

AprilYUZhang:
> > > the update of the option names of `tinyhouse` is more difficult than that of `alphapeel` as the option names are used everywhere among the 10 files of `tinyhouse`. It would be more likely for us to miss some of them while modifying the code.
> > 
> > 
> > @XingerTang why would this be hard? It's mostly an input and maybe some variable name change, which is less than doing extra logic, particularly in [5821d43](https://github.com/AlphaGenes/AlphaPeel/commit/5821d432e533277590a2b620dfbb4d8feb76708e). The part that I am worried is that while these current changes are not large, they are adding extra logic to `AlphaPeel` and when we decide to polish `tinyhouse` we will forget that some of this extra logic should be removed from `AlphaPeel`.
> 
> @gregorgorjanc It would be hard doesn't because it's hard to write anything, but the code of `AlphaPeel` and `tinyhouse` are messy, and it is hard to track where the option names go. For those option names that are adding/modifying functionality, it is not enough to do "search and replace" as we need to find everywhere it is used and understand how it is used. In the above, @AprilYUZhang found where the errors are updated and modified the line of the condition for estimating genotype error and sequence error before that line, from
> 
> ```
>         if args.esterrors:
>             PeelingUpdates.updatePenetrance(pedigree, peelingInfo)
> ```
> 
> to
> 
> ```
>         if args.est_geno_error_prob or args.est_seq_error_prob:
>             PeelingUpdates.updatePenetrance(pedigree, peelingInfo)
> ```
> 
> but that is not enough, we need to look up quite closely to see which line corresponds to the update of genotype error and which line corresponds to the update of sequence error inside `PeelingUpdates.updatePenetrance(pedigree, peelingInfo)` and add additional conditions for them, even though there is no `args.esterrors` in it. We need to be quite familiar with the code to do all the changes without missing anything. And, when we are facing `tinyhouse`, there are 10 files. It is easy to lose track of where and how some values are passed from one place to another, but we need to understand the uses of all of them before making changes. That is the time-consuming part. Also, we don't want code inside `tinyhouse` inconsistent, so we cannot modify only the `AlphaPeel` related part of `tinyhouse`.
> 
> When we polish `tinyhouse`, we also have to check every place of `AlphaPeel` that is related to `tinyhouse` to avoid inconsistency, so we won't miss any extra logic.

Thank you! you explain so clearly.

src/tinypeel/tinypeel.py
```diff
@@ -213,6 +213,107 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+
+def get_input_options():
+    parse_dictionary = dict()
+    parse_dictionary["bfile"] = lambda parser: parser.add_argument(
+        "-plink_file",
+        default=None,
+        required=False,
+        type=str,
+        nargs="*",
+        help="A file in plink (binary) format. Only stable on Linux).",
```
gregorgorjanc, line 250: @AprilYUZhang remove last `)`

src/tinypeel/tinypeel.py
```diff
@@ -213,6 +213,107 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+    parse_dictionary["genotypes"] = lambda parser: parser.add_argument(
+        "-geno_file",
+        default=None,
+        required=False,
+        type=str,
+        nargs="*",
+        help="A file in AlphaGenes format.",
+    )
+    parse_dictionary["reference"] = lambda parser: parser.add_argument(
+        "-reference",
```
gregorgorjanc, line 261: @AprilYUZhang do we mention this file at all in documentation? I think we are not. Please open an issue to sort this.

src/tinypeel/tinypeel.py
```diff
@@ -213,6 +213,107 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+    parse_dictionary["pedigree"] = lambda parser: parser.add_argument(
+        "-ped_file",
+        default=None,
+        required=False,
+        type=str,
+        nargs="*",
+        help="A pedigree file in AlphaGenes format.",
+    )
+    parse_dictionary["phasefile"] = lambda parser: parser.add_argument(
+        "-phasefile",
```
gregorgorjanc, line 285: @AprilYUZhang we will call this file diffrently!?

src/tinypeel/tinypeel.py
```diff
@@ -213,6 +213,107 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+        default=None,
+        required=False,
+        type=str,
+        nargs="*",
+        help="A phase file in AlphaGenes format.",
+    )
+    parse_dictionary["startsnp"] = lambda parser: parser.add_argument(
+        "-startsnp",
```
gregorgorjanc, line 293: @AprilYUZhang should not this be `-start_snp`?

src/tinypeel/tinypeel.py
```diff
@@ -213,6 +213,107 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+        default=None,
+        required=False,
+        type=int,
+        help="The first marker to consider. The first marker in the file is marker '1'. Default: 1.",
+    )
+    parse_dictionary["stopsnp"] = lambda parser: parser.add_argument(
+        "-stopsnp",
```
gregorgorjanc, line 300: @AprilYUZhang should not this be `-stop_snp`?

gregorgorjanc (changes requested):
@AprilYUZhang I reviewed your recent changes and requested some modifications.

src/tinypeel/tinypeel.py
AprilYUZhang: Yeah, but I will open a new commit to fix it.

AprilYUZhang:
I find ".called." is not fixed in `writeBinaryCalledGenotypes`. Although it hasn’t been realized, do we need to change it? @gregorgorjanc @XingerTang

XingerTang:
> I find ".called." is not fixed in `writeBinaryCalledGenotypes`. Although it hasn’t been realized, do we need to change it? @gregorgorjanc @XingerTang

@AprilYUZhang It's not important for now. I prefer to leave that to the future.

gregorgorjanc:
> > I find ".called." is not fixed in `writeBinaryCalledGenotypes`. Although it hasn’t been realized, do we need to change it? @gregorgorjanc @XingerTang
> 
> @AprilYUZhang It's not important for now. I prefer to leave that to the future.

@AprilYUZhang open an issue and we get back to it in future

AprilYUZhang:
I updated the summary of changes #113

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -66,47 +66,59 @@ def readInSeg(pedigree, fileName, start=None, stop=None):
...
-    np.savetxt(args.out + ".genoError", peelingInfo.genoError, fmt="%f")
-    np.savetxt(args.out + ".seqError", peelingInfo.seqError, fmt="%f")
-    # np.savetxt(args.out + ".trans", peelingInfo.transmissionRate, fmt = "%f")
-    np.savetxt(args.out + ".maf", peelingInfo.maf, fmt="%f")
+    np.savetxt(args.out_file + ".geno_error_prob.txt", peelingInfo.genoError, fmt="%f")
+    np.savetxt(args.out_file + ".seq_error_prob.txt", peelingInfo.seqError, fmt="%f")
+    np.savetxt(
+        args.out_file + ".rec_prob.txt", np.empty((1, 1)), fmt="%f"
+    )  # not be realized, just as a placeholder
+    # np.savetxt(args.out_file + ".trans", peelingInfo.transmissionRate, fmt = "%f")
```
gregorgorjanc, line 74: @AprilYUZhang do you know what exactly is in `peelingInfo.transmissionRate`? I wonder if this anyway connected to mutations? Open an issue and we can look at it later.

src/tinypeel/tinypeel.py
```diff
@@ -213,19 +213,166 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+    )
+    parse_dictionary["seq_error_prob"] = lambda parser: parser.add_argument(
+        "-seq_error_prob",
+        default=0.001,
+        required=False,
+        type=float,
+        help="Sequencing error rate. Default: 0.001.",
+    )
+    parse_dictionary["recombination"] = lambda parser: parser.add_argument(
+        "-recomb",
```
gregorgorjanc, line 233: @AprilYUZhang should this be `rec_length`?

src/tinypeel/tinypeel.py
```diff
@@ -213,19 +213,166 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+
+def get_output_options():
+    parse_dictionary = dict()
+
+    parse_dictionary["writekey"] = lambda parser: parser.add_argument(
+        "-out_id_order",
+        default="id",
+        required=False,
+        type=str,
+        help='Determines the order in which individuals are ordered in the output file based on their order in the corresponding input file. Animals not in the input file are placed at the end of the file and sorted in alphanumeric order. These animals can be surpressed with the "-onlykeyed" option. Options: id, pedigree, genotypes, sequence, segregation. Defualt: id.',
```
gregorgorjanc, line 325: Change `-onlykeyed` to `-out_id_only`

src/tinypeel/tinypeel.py
```diff
@@ -213,19 +213,166 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+    )
+    parse_dictionary["onlykeyed"] = lambda parser: parser.add_argument(
+        "-out_id_only",
+        action="store_true",
+        required=False,
+        help='Flag to surpress the animals who are not present in the file used with -outputkey. Also surpresses "dummy" animals.',
```
gregorgorjanc, line 331: Change `outputkey` to `???`

src/tinypeel/Peeling/PeelingIO.py
AprilYUZhang:
```
def setupTransmission(length, peelingInfo):
    if peelingInfo.positions is None:
        localMap = np.linspace(0, 1, num=peelingInfo.nLoci, dtype=np.float32)
    else:
        localMap = (
            peelingInfo.positions / peelingInfo.positions[-1]
        )  # This should be sorted. Need to add in code to check.
    for i in range(peelingInfo.nLoci - 1):
        distance = localMap[i + 1] - localMap[i]
        distance = distance * length
        peelingInfo.transmissionRate[i] = distance
```
I found the code that calculates the distance between a locus and an adjacent locus. It seems to be used in the updateSeg(), but now it has been abandoned.
```
# def updateSeg(peelingInfo):
#     # Idea: Split the chromosome into ~10 blocks with slightly different starts/stops. Estimate at each one and then sum.
#     # Not sure if this is the best way, but probably will work okay.

#     distances = [
#         estDistanceFromBreaks(0, peelingInfo.nLoci - 1, nBreaks, peelingInfo)
#         for nBreaks in [5, 10, 20]
#     ]
#     distance = np.mean(distances[0:2])
#     print(distances, ":", distance)

#     setupTransmission(distance, peelingInfo)
```

src/tinypeel/tinypeel.py
```diff
@@ -213,19 +213,166 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+        help="Number of threads to use for io. Default: 1.",
+    )
+
+    return parse_dictionary
+
+
+def get_multithread_options():
+    parse_dictionary = dict()
+    parse_dictionary["iothreads"] = lambda parser: parser.add_argument(
+        "-n_io_threads",
```
gregorgorjanc, line 347: @AprilYUZhang please ensure that documentation files are updated too - I don't think I have spotted this option being documented and I did not pay attention before either.

gregorgorjanc (changes requested):
@AprilYUZhang minor changes/revisions requested

src/tinypeel/tinypeel.py
gregorgorjanc: @AprilYUZhang hmm, the help text is the same as for `rec_length` as is the default suggesting we are confusing something here. At any rate, looking at the table https://github.com/AlphaGenes/AlphaPeel/issues/113 we have `rec_length` as input for the genetic length of the chromosome, `est_rec_prob` to flag estimation of the recombination rate between loci, and `.rec_prob` for file output (I made this clear in https://github.com/AlphaGenes/AlphaPeel/issues/113). There is no `recombination` option that I am aware of, but let me know if I am confusing something.

https://github.com/AlphaGenes/AlphaPeel/issues/97
https://github.com/AlphaGenes/AlphaPeel/issues/104

AprilYUZhang:
the problem of these two options are similar toIn  commit #[0632f47](https://github.com/AlphaGenes/AlphaPeel/commit/0632f47a3492b06a9f75491f08136c41c61146ce). I just changed the part parsing argument rather than all. When we consider to change tinyhouse, it will be fixed in the whole software. 

> In commit # [0632f47](https://github.com/AlphaGenes/AlphaPeel/commit/0632f47a3492b06a9f75491f08136c41c61146ce), I copied the `get_input_options()` from tinyhouse, but I just changed the input name when it parsed parameters. The reason why I didn't change the name in the `parse_dictionary`, `pedigree, genotypes, seqfile, bfile` are involved many times in the program calls of tinyhouse.

src/tinypeel/tinypeel.py
AprilYUZhang: I'm not sure if you pointed at `writekey`, it has been changed to `out_id_order`.

gregorgorjanc:
> the problem of these two options are similar toIn commit #[0632f47](https://github.com/AlphaGenes/AlphaPeel/commit/0632f47a3492b06a9f75491f08136c41c61146ce). I just changed the part parsing argument rather than all. When we consider to change tinyhouse, it will be fixed in the whole software.
> 
> > In commit # [0632f47](https://github.com/AlphaGenes/AlphaPeel/commit/0632f47a3492b06a9f75491f08136c41c61146ce), I copied the `get_input_options()` from tinyhouse, but I just changed the input name when it parsed parameters. The reason why I didn't change the name in the `parse_dictionary`, `pedigree, genotypes, seqfile, bfile` are involved many times in the program calls of tinyhouse.

@AprilYUZhang I don't know to which question/comment you are referring to with this answer. Can you clarity? Is this related to my comment in this thread https://github.com/AlphaGenes/AlphaPeel/pull/122#discussion_r1399369654?

src/tinypeel/Peeling/PeelingIO.py
gregorgorjanc: @AprilYUZhang the above help flag has still `-outputkey` mentioned, while it should be `out_id_order`

src/tinypeel/tinypeel.py
AprilYUZhang: It should be "-out_id_order". In the documents, it is right.

src/tinypeel/tinypeel.py
AprilYUZhang: I get your means; I made the mistake that rec_prob is not an option rather than an output file name.

gregorgorjanc:
@AprilYUZhang I think this looks very good all in all! Are there any more planned changes - it seems you have done all of the points in https://github.com/AlphaGenes/AlphaPeel/issues/113!? 

@XingerTang I would appreciate quick code review from your end before we merge this in.

src/tinypeel/tinypeel.py
```diff
@@ -213,19 +213,160 @@ def generateSingleLocusSegregation(peelingInfo, pedigree, args):
...
+    )
+    parse_dictionary["seed"] = lambda parser: parser.add_argument(
+        "-seed",
+        default=None,
+        required=False,
+        type=int,
+        help="A random seed to use for debugging.",
+    )
+
+    return parse_dictionary
```
XingerTang, line 308: @AprilYUZhang 
Even though you keep the keys of the `parse_dictionary` unchanged, you still need to change how you access the values via `args` inside the main program. 

I did a test program, it basically does the same thing as those functions in `tinyhouse.InputOutput`

```python
import argparse
import sys

parser = argparse.ArgumentParser(description = 'Test')

def add_arguments_from_dictionary(parser, arg_dict, options = None):
    if options is None:
        for key, value in arg_dict.items():
            value(parser)
    else:
        for option in options:
            if option in arg_dict:
                arg_dict[option](parser)
            else:
                print("Option not found:", option, arg_dict)

def get_input_options():

    parse_dictionary = dict()
    parse_dictionary["bfile"] = lambda parser: parser.add_argument(
        '-plink_file',   
        default=None, 
        required=False, 
        type=str, 
        nargs="*", 
        help='A file in plink (binary) format. Only stable on Linux).'
        )
    
    return parse_dictionary

input_parser = parser.add_argument_group("Input Options")
add_arguments_from_dictionary(
        input_parser,
        get_input_options(),
        options=[
            "bfile"
        ],
)

args = sys.argv[1:]
args = parser.parse_args(args)
print(f"{args.plink_file=}")
print(f"{args.bfile=}")
```
And I get 
```bash
(base) evie@Xingers-MacBook-Air test % python /Users/evie/test/src/test/testing.py -plink_file "FILE NAME"
args.plink_file=['FILE NAME']
Traceback (most recent call last):
  File "/Users/evie/test/src/test/testing.py", line 43, in <module>
    print(f"{args.bfile=}")
AttributeError: 'Namespace' object has no attribute 'bfile'
```
So the `args` can only know `plink_file` as the name in the command line `plink_file` but not the dictionary key `bfile`.

You still need to change the places where the option names are changed in the main program. I think one way to do this is by redefining the attributes of the `args` 
```python
args = sys.argv[1:]
args = parser.parse_args(args)
args.bfile = args.plink_file
print(f"{args.bfile=}")
```
And with this change I can get
```bash
(base) evie@Xingers-MacBook-Air test % python /Users/evie/test/src/test/testing.py -plink_file "FILE NAME"
args.bfile=['FILE NAME']
```

src/tinypeel/Peeling/PeelingUpdates.py
XingerTang, line 1: @AprilYUZhang Please add conditions of `if args.est_geno_error_prob:` and `if args.est_seq_error_prob:` before all the genotype error updates and the sequence error updates respectively

XingerTang (changes requested):

XingerTang:
@AprilYUZhang @gregorgorjanc There are still some changes that need to be made before we merge. Please see the above comments.

XingerTang:
@AprilYUZhang @gregorgorjanc I just noticed that we still have an `s` in `n_cycles` which is the updated name of `ncycles`. Should we keep it or change it to `n_cycle`?

gregorgorjanc:
> @AprilYUZhang @gregorgorjanc I just noticed that we still have an `s` in `n_cycles` which is the updated name of `ncycles`. Should we keep it or change it to `n_cycle`?

@XingerTang well spotted! @AprilYUZhang please check all the options and let us know what you suggest as a unified approach regarding the plural and the associated “s”.

src/tinypeel/tinypeel.py
AprilYUZhang: Maybe today we can talk about it, but why you need use args.bfile in the test? I'm a little confused. Because I don't know what is wrong in the main programs.

src/tinypeel/Peeling/PeelingUpdates.py
AprilYUZhang: I get it

src/tinypeel/tinypeel.py
XingerTang: @AprilYUZhang We don't need `args.bfile` in the test for now, I just use it as an example to explain why I think you need to add a few more lines to make the code work.

XingerTang (approved):

AprilYUZhang:
Solved the above problems.

XingerTang:
@gregorgorjanc 
Could you rerun this following workflow
Screen Shot 2023-11-24 at 8 51 05 PM
By clicking the job bar on the left
Screen Shot 2023-11-24 at 8 51 29 PM

XingerTang:
@gregorgorjanc @AprilYUZhang 
Sorry I forgot the commit here is not up to date and sent the wrong comment, I already deleted that. 
@AprilYUZhang Could you pull the recent changes that just merged into the branch, and let's see how the test go

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -66,47 +66,59 @@ def readInSeg(pedigree, fileName, start=None, stop=None):
...
+    np.savetxt(
+        args.out_file + ".alt_allele_prob.txt", peelingInfo.alt_allele_prob, fmt="%f"
+    )
```
XingerTang, line 77: Please change `peelingInfo.alt_allele_prob` back to `peelingInfo.maf`.

src/tinypeel/tinypeel.py
```diff
@@ -416,28 +562,45 @@ def getArgs():
...
+    args.phasefile = args.hap_file
+    args.seqfile = args.seq_file
+    args.pedigree = args.ped_file
+    args.startsnp = args.start_snp
+    args.stopsnp = args.stop_snp
+    args.writekey = args.out_id_order
+    args.onlykeyed = args.out_id_only
+    args.iothreads = args.n_io_thread
+    args.maxthreads = args.n_thread
+
```
XingerTang, line 577: Please add a line  of `args.segfile = args.seg_file`.

AprilYUZhang: ok

src/tinypeel/tinypeel.py
```diff
@@ -28,8 +28,7 @@ def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
         #     print("Estimating the transmission rate is currently a disabled option")
         # PeelingUpdates.updateSeg(peelingInfo) #Option currently disabled.
 
-        if args.esterrors:
-            PeelingUpdates.updatePenetrance(pedigree, peelingInfo)
+        PeelingUpdates.updatePenetrance(pedigree, peelingInfo, args)
```
XingerTang, line 31: Please add the condition `if args.est_geno_error_prob or args.est_seq_error_prob:` back

src/tinypeel/tinypeel.py
```diff
@@ -416,28 +562,46 @@ def getArgs():
 
 def main():
     args = getArgs()
```
XingerTang, line 564: Please also add

```
    if args.start_snp:
            args.start_snp -= 1
    if args.stop_snp:
            args.stop_snp -= 1
```

This is modified from `tinyhouse.InputOutput.parseArgs`, but now we also need it here.

XingerTang (changes requested):

XingerTang:
@gregorgorjanc Could you please merge this pull request?

gregorgorjanc:
> @gregorgorjanc Could you please merge this pull request?

@XingerTang i lost track of which of the above discussions and requested changes have been addressed l please go over them and resolve each one (maybe the changes have all been achieved?) and then we merge this in,

XingerTang:
Yes, all the changes are achieved and every problem is addressed. I think we can merge this in.

Issue #153, opened 2024-01-05
Provide map file for non-hybrid mode


XingerTang:
See discussion in #144.

PR #154, opened 2024-01-09
Provide map file input option for non-hybrid mode #153

XingerTang:
#153


src/tinypeel/Peeling/PeelingInfo.py
```diff
@@ -102,8 +109,8 @@ def setupTransmission(length, peelingInfo):
     if peelingInfo.positions is None:
         localMap = np.linspace(0, 1, num=peelingInfo.nLoci, dtype=np.float32)
     else:
-        localMap = (
-            peelingInfo.positions / peelingInfo.positions[-1]
+        localMap = (peelingInfo.positions - peelingInfo.positions[0]) / (
+            peelingInfo.positions[-1] - peelingInfo.positions[0]
         )  # This should be sorted. Need to add in code to check.
```
gregorgorjanc, line 114: @XingerTang do we know if `InputOutput.readMapFile()` checks for sorted output or sorts the input? Please open an issue.

tests/functional_tests/test_map_input/ped_file.txt
gregorgorjanc, line 1: @XingerTang you have not listed parents as `M*` or `F*` but just `*`. I assume `1` should be `M1` etc.?

src/tinypeel/Peeling/PeelingInfo.py
XingerTang: @gregorgorjanc `InputOutput.readMapFile()` does not check, we could implement it but it would be quite expensive to actually write a sorting algorithm (a quick research gives the minimum would be O(nlog(n))).

tests/functional_tests/test_map_input/ped_file.txt
XingerTang: @gregorgorjanc The ids in ped_file are `M*` or `F*`. The one that starts with `*` is the `map_file`, where the columns represent the chromosome number, marker name, and base pair position respectively.

src/tinypeel/Peeling/PeelingInfo.py
gregorgorjanc: @XingerTang also then we should add a test where you scramble the order of rows in the map file and correspondingly the columns of the genotype file. The question now is what should be the output - in the sorted order or the original order (another option, doh)?

But, should some of this functionality be in `tinyhouse` or in `AlphaPeel`?

Best to move this discussion into an issue!

XingerTang: @gregorgorjanc All of these functionalities are in `tinyhouse`.

src/tinypeel/Peeling/PeelingInfo.py
```diff
@@ -102,8 +109,8 @@ def setupTransmission(length, peelingInfo):
     if peelingInfo.positions is None:
         localMap = np.linspace(0, 1, num=peelingInfo.nLoci, dtype=np.float32)
     else:
-        localMap = (
-            peelingInfo.positions / peelingInfo.positions[-1]
+        localMap = (peelingInfo.positions - peelingInfo.positions[0]) / (
```
gregorgorjanc, line 112: @XingerTang nice catch!!!!

gregorgorjanc (changes requested):
@XingerTang minor changes requested;)

src/tinypeel/Peeling/PeelingInfo.py
gregorgorjanc: > @gregorgorjanc `InputOutput.readMapFile()` does not check, we could implement it but it would be quite expensive to actually write a sorting algorithm (a quick research gives the minimum would be O(nlog(n))).

I know! Well, we have the issue now open at https://github.com/AlphaGenes/AlphaPeel/issues/155 so I propose we remove the comment in row 114 since this test/check should be done elsewhere and that comment will be redundant and in the wrong place anyway.

gregorgorjanc: > @gregorgorjanc The issue is opened. But I think we don't have to be too worried about the genotype input, it has an option for the output order,`out_id_order`, that allows the output to be in the order of ids or the sequence of the input of different files.

That is for the order of rows. Map file positions deal with columns. Anyway, let's drop this discussion for later in https://github.com/AlphaGenes/AlphaPeel/issues/155

XingerTang: > > @gregorgorjanc `InputOutput.readMapFile()` does not check, we could implement it but it would be quite expensive to actually write a sorting algorithm (a quick research gives the minimum would be O(nlog(n))).
> 
> I know! Well, we have the issue now open at #155 so I propose we remove the comment in row 114 since this test/check should be done elsewhere and that comment will be redundant and in the wrong place anyway.

Sure!

tests/functional_tests/test_map_input/ped_file.txt
XingerTang: @gregorgorjanc That's a mistake, I will correct it.

gregorgorjanc (approved):

gregorgorjanc:
@XingerTang one final request - we should list changes in the NEWS/CHANGELOG file. Please do this - I hope to remember for every merged PR from now onwards;) It is easier this way than later going over the list of commits;) Thanks for all these contributions.

XingerTang:
@gregorgorjanc Sure!

XingerTang:
@gregorgorjanc Can you create a new branch `feat_recomb_new_changelog` so that I can open a separate pull request for the log files? Thanks!

gregorgorjanc:
> @gregorgorjanc Can you create a new branch `feat_recomb_new_changelog` so that I can open a separate pull request for the log files? Thanks!

@XingerTang done at https://github.com/AlphaGenes/AlphaPeel/tree/feat_recomb_new_changelog

gregorgorjanc:
@XingerTang thanks for squashing the commit and for this PR! Merging this in!

Issue #142, opened 2023-12-11
Implement alt_allele_prob option


AprilYUZhang:
So we can change allele freq in the founders (not setting to 0.5 and not estimating from the data)

RosCraddock:
@XingerTang - assigned to this issue.

XingerTang:
@RosCraddock @gregorgorjanc 

Coding tasks to do:
- [ ] Modification in `tinyhouse.pedigree` to store the information of the metafounders while reading in the pedigree file
  - [ ] Add `MetaFounder` flag/attribute to `Individual` class, while reading in the pedigree data, set the flag to `True` for each individual with their names start with `MF_` (can also check if metafounders are actually founders, raise errors if not)
  - [ ] Add automatically generated `MF_1` individual to the individual list of the `Pedigree` object after the input pedigree data is read
  - [ ] Modify `Pedigree.readInPedigree` so that for each individual in the pedigree
  ```
  if both parents == None None and not MetaFounder:
    set MF_1 as parents
  ```
- [ ] Modification in `alphapeel.peelinginfo` to store the corresponding alternative allele frequency for each of the metafounders
  - [ ] Add `jit_peelingInformation.nMF` to store the number of metafounders in the population
  - [ ] Add `jit_peelingInformation.MFList` to store the list of the metafounder individuals (or their ids)
  - [ ] Initialize `jit_peelingInformation.maf` as an `nMF` $\times$ `nLoci` numpy matrix
- [ ] Modification in `alphapeel.peelinginfo` to allow user-defined alternative allele frequency to be used in the calculation
  - [ ] Calculate the values based on the provided alternative allele frequency while initializing the anterior probabilities of the metafounders
- [ ] Modification in `alphapeel.tinypeel` to add the option `alt_allele_prob`
  - [ ] Add option `alt_allele_prob`
  - [ ] Add preference for the case when both `alt_allele_prob` and `est_alt_prob` are used

gregorgorjanc:
I spoke with @AprilYUZhang today and she pointed out some bits that I wan't to clarify here.

Current state in AlphaPeel is the following:
1. Estimate alternative allele probability based on observed genotyped data (using Newton method on genotype probabilities from observed genotypes (something like this), so accounting for genotyping error)
2. Take 1. and set it for the rest of the program execution
3. Use 1. to set anterior term for founders (since alternative allele probability is fixed, so are these anterior terms)
[...]

PR #175, opened 2024-09-04
Metafounder updates

RosCraddock:
Completed updates for metafounder implementation (links to issue #142 ). There is more detail in the commits, but just for an overview:

1. Added two input parameters: alt_allele_prob_file and main_metafounder
2. Code to update the anterior of founders based on user-inputted alternative allele frequency.
3. New function for outputting the alternative allele frequency when alt_allele_prob_file is **not** inputted.
4. Minor updates to Usage.file

Please review when you can and let me know if you have any suggestions/changes @XingerTang


src/tinypeel/Peeling/PeelingIO.py
XingerTang, line 1: For the function `writeOutAltAlleleProb()`, we don't have to iterate through the whole pedigree, the anterior of the founders are never updated during the peeling, we can just read the entries in `pedigree.AAP`. 
As for the later estimation of the alternative allele probabilities, that function would automatically generate the alternative allele probabilities for each metafounder, which can be stored in the `pedigree.AAP` directly.

src/tinypeel/tinypeel.py
```diff
@@ -17,7 +17,14 @@ def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
     # Right now maf _only_ uses the penetrance so can be estimated once.
     if args.est_alt_allele_prob:
         PeelingUpdates.updateMaf(pedigree, peelingInfo)
-
+    if args.alt_allele_prob_file is not None:
+        pedigree.readInAAP(pedigree, args.alt_allele_prob_file)
+        for ind in pedigree:
+            if ind.MetaFounder is not None:
```
XingerTang, line 23: A condition of the individual to be the founder is also required (try `ind.isFounder()`). In the `tinyhouse.Pedigree`, I set metafounders for all of the individuals for simplicity, so the progenies of the progenies of different metafounders all share a default metafounder, so we don't want to update their anterior probabilities otherwise it would interfere with the information while passing down the pedigree if the alternative allele probability of the default metafounder is not 0.5.

src/tinypeel/tinypeel.py
```diff
@@ -383,11 +403,12 @@ def getArgs():
             "phasefile",
             "seqfile",
             "pedigree",
+            "altalleleprobfile",
             "startsnp",
             "stopsnp",
+            "mainmetafounder"
```
XingerTang, line 409: Please use `alt_allele_prob_file` and `main_meta_founder`, these two actually correspond to the user input.

src/tinypeel/tinypeel.py
```diff
@@ -573,6 +594,8 @@ def main():
     args.pedigree = args.ped_file
     args.startsnp = args.start_snp
     args.stopsnp = args.stop_snp
+    args.altalleleprobfile = args.alt_allele_prob_file
+    args.mainmetafounder = args.main_metafounder
```
XingerTang, line 598: These two can be removed. I know these lines of code are a bit confusing, but they are just a temporary solution to the mismatch of the argument names in AlphaPeel and tinyhouse. As long as all of our operations are all in AlphaPeel, we don't need this.

src/tinypeel/tinypeel.py
```diff
@@ -602,11 +625,12 @@ def main():
     )
     if not args.no_param:
         PeelingIO.writeOutParamaters(peelingInfo)
+    if not args.altalleleprobfile: # Only write alternative allele probabilities if user has **not** inputted them
+        PeelingIO.writeOutAltAlleleProb(pedigree, peelingInfo)
```
XingerTang, line 629: I'm a bit confused, this output file should be output whenever user use the option `alt_allele_prob`. We need the output file when the estimation of the alternative allele frequency is used.

XingerTang (changes requested):
@RosCraddock Thank you for your work! Please check some of the comments I make.

src/tinypeel/Peeling/PeelingIO.py
RosCraddock: Ok, I can see that would be more efficient. For that, will I need to add code to save the default alternative allele probability (0.5) to each metafounder in `pedigree.AAP`?

src/tinypeel/tinypeel.py
RosCraddock: Ahh, I had missed that bit. Thank you!

src/tinypeel/tinypeel.py
RosCraddock: Will do. Metafounder is one word so would it be preferable to use main_metafounder?

src/tinypeel/tinypeel.py
RosCraddock: This code will output the file when the default or estimated alternative allele probability is used. It will not produce an output file if the user has defined the alternative allele probability in the input. I thought it would save an additional command for the user. But, happy to add the option instead!

src/tinypeel/Peeling/PeelingIO.py
XingerTang: It's OK, I already did that.

src/tinypeel/tinypeel.py
XingerTang: AlphaPeel is working in a way that the users control all the additional output files, so I think it is better to keep it.

src/tinypeel/tinypeel.py
```diff
@@ -17,7 +17,14 @@ def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
     # Right now maf _only_ uses the penetrance so can be estimated once.
     if args.est_alt_allele_prob:
         PeelingUpdates.updateMaf(pedigree, peelingInfo)
-
+    if args.alt_allele_prob_file is not None:
+        pedigree.readInAAP(pedigree, args.alt_allele_prob_file)
```
RosCraddock, line 21: I think I can remove this line (line 21) as it is already triggered in tinyhouse/InputOutPut Line 237 to 239 (https://github.com/AlphaGenes/tinyhouse/pull/15/commits/a276aab24976c12d24524a192ab8aaf109bfb147). Is this correct @XingerTang?

src/tinypeel/Peeling/PeelingIO.py
RosCraddock: @XingerTang Could you share where in the code you did? Sorry, I can't find it!

XingerTang: @RosCraddock 
Screen Shot 2024-09-05 at 4 28 41 PM

src/tinypeel/tinypeel.py
XingerTang: @RosCraddock You are right.

src/tinypeel/Peeling/PeelingIO.py
RosCraddock: Is this only called if the user inputs an alt_allele_prob_file?

I am thinking of the case where the user does not input an alt_allele_prob_file, nor requests est_alt_allele_prob, but asks for the alt_allele_prob output. This would only result in an empty outputted alt_allele_prob.txt file since here pedigree.AAP would be empty. Nothing else would be impacted, so it is only a minor issue, and the user is unlikely to do this if they understand how AlphaPeel/ pedigree-based inference works.

XingerTang: @RosCraddock Yes, it is only called if the user inputs the file.
For the case you mentioned, it is actually reasonable to have an empty output, as we don't know the range of the individuals that need to be considered to calculate the alternative allele probabilities, and at the same time, the best way to calculate the alternative allele probabilities should be the method we implement for the estimation algorithm. This specific case basically requires us to do the estimation without using the estimated result, which can be done, but better to be implemented elsewhere.

RosCraddock: In the above example the alternative allele frequency used for each metafounder will be 0.5 (i.e the default). Although I see it would be reasonable to be empty, it would be more useful to list 0.5 for all defined metafounders (per the pedigree) since it is what is used in the model. 
I have added some code to update the pedigree.AAP with the default (so 0.5 
 for nLoci) for each defined metafounder prior to peeling. This may also be useful later in the updateMaf function for estimating the alternative allele frequency (depending on how we do this).

XingerTang: @RosCraddock That's a good way to do it. What I was thinking is the user may want to obtain more genetic information. It would be better if we could estimate the alternative allele probability in the most accurate way after the peeling is done.

src/tinypeel/tinypeel.py
```diff
@@ -12,19 +12,23 @@
...
     if args.alt_allele_prob_file is not None:
-        pedigree.readInAAP(pedigree, args.alt_allele_prob_file)
         for ind in pedigree:
-            if ind.MetaFounder is not None:
+            if ind.isFounder() and ind.MetaFounder is not None:
                 mfx = ind.MetaFounder
                 aaf = pedigree.AAP[mfx]
                 aafGeno = ProbMath.getGenotypesFromMaf(aaf)
                 peelingInfo.anterior[ind.idn, :, :] = aafGeno
+    else:
```
XingerTang, line 24: There are still possibilities that the input alternative allele probabilities are incomplete, that is, some of the metafounders are mentioned in the pedigree file while not being specified in the alternative allele probability input file. These metafounders won't have the default alternative allele probabilities with this if-else condition.

XingerTang (changes requested):
A minor issue, please see the comment.

I personally still tend to not introduce default alternative allele probability to other metafounders, as it would increase the number of the anterior probabilities that need to be updated while it won't make any difference before and after the update because the initialized anterior probabilities are also generated based on the 0.5 alternative allele probabilities.

XingerTang:
@RosCraddock I noticed that the pre-commit check is not passed, if you haven't installed/used the pre-commit before, we can schedule a meeting for this.

gregorgorjanc (approved):
LGTM

RosCraddock:
@gregorgorjanc Apologies for all the commits, I will squash them all shortly. Once the [tinyhouse pull request](https://github.com/AlphaGenes/tinyhouse/pull/17) has been merged, this should pass the GitHub action checks. Then, I believe testing is complete for the metafounders (**without** estimation of the alternative allele frequency).

gregorgorjanc:
No worries @RosCraddock! I have dealt with that PR. Let's see how action checks work and then move onwards. Please remind me once actions work;)

XingerTang:
@RosCraddock Despite the exit code, everything is very good! The exit code is not very informative regarding which specific errors have been raised, ideally, we would want to assess the error message. I know it's not easy to implement, but maybe we can try.

XingerTang:
@RosCraddock Thank you for the effort, I can see it won't be easy... Could you try [`capsys()`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.capture.capsys) or/and the `with pytest.raises(<Error Type>, match = <Error Message>)`? If this still doesn't work then we can skip this part.

RosCraddock:
@XingerTang I have quickly tried both with no success unfortunately!

XingerTang:
> @XingerTang I have quickly tried both with no success unfortunately!

@RosCraddock OK, leave it there for now then. Maybe open an issue to come back later?

gregorgorjanc:
@RosCraddock what is the next action with this PR?

RosCraddock:
> @RosCraddock what is the next action with this PR?

@gregorgorjanc I noticed another bug yesterday in Tinyhouse that I have now resolved (see [pull request 18](https://github.com/AlphaGenes/tinyhouse/pull/18)). Once that is reviewed and merged, I can update the submodule here. Then, this PR can be merged. However, the estimation of the alternative allele frequency with metafounders is not yet functional. When the code is ready, I can start a new PR for that. I am still doing some testing locally - will add to the issue later today!

gregorgorjanc (approved):
LGTM!

src/tinypeel/tinypeel.py
gregorgorjanc: @RosCraddock thoughts on this comment?

gregorgorjanc: @RosCraddock do you need to add a test case for this?

gregorgorjanc:
@RosCraddock I had a look at the changes. Most of it looks ok. There is the unresolved question on what to do when a user skips one MF in the provided file compared to a pedigree AND when a user provides a MF in the provided file that is not in a pedigree. We could just throw errors when either of that happens and ping it back to a user!

@XingerTang any additional thoughts?

src/tinypeel/tinypeel.py
RosCraddock: Oh, I had forgotten about this comment. I may be able to redo this to first check whether pedigree.AAP[mfx] has a value yet before assigning a default. If not, I can add an error instead. Yes, I will add a test case for this. @gregorgorjanc

RosCraddock:
> @RosCraddock I had a look at the changes. Most of it looks ok. There is the unresolved question on what to do when a user skips one MF in the provided file compared to a pedigree AND when a user provides a MF in the provided file that is not in a pedigree. We could just throw errors when either of that happens and ping it back to a user!
> 
> @XingerTang any additional thoughts?

I previously added a conditional statement to assign the default where a MF is not listed in the alt allele prob file but is in the pedigree. I will add a test case for this. I did not consider the other scenario, but I think the addition would be ignored in the peeling but listed in the alt allele prob output (if requested). Again, I will add a test case for this. @gregorgorjanc

src/tinypeel/tinypeel.py
gregorgorjanc: Perfect! Will leave this conversation open as a reminder to add the test case.

XingerTang:
All look to me :)

RosCraddock:
@gregorgorjanc I agree, I will go ahead with 3 with an addition of a warning to inform the user!

gregorgorjanc (approved):
LGTM! but would appreciate your thoughts as well @XingerTang !

gregorgorjanc:
@RosCraddock we mentioned to "warn" about a deleted non-existent MF in pedigree. Do we have any established way we report such warnings?

RosCraddock:
Apologies, I forgot to add the warning. I have just pushed that through now. I am not sure it is an established way, but it seems that these warnings are just printed in the terminal prior to the peeling commencing.

gregorgorjanc (approved):
LGTM!

XingerTang:
@RosCraddock `tinyhouse.InputOutput` are using `warning.warn` to raise the warnings, maybe we can also use it?

RosCraddock:
@XingerTang Ahh, ok. I did not check what happened in tinyhouse. I will implement the warning.warn. I think there are some other parts in AlphaPeel that could be updated with that too so will create an issue as well.

src/tinypeel/tinypeel.py
gregorgorjanc: @XingerTang @RosCraddock I am going over this PR discussion trying to bring it to a merge, but on this part I am a bit confused what is going on. Can you please help me understand this:

>  n the tinyhouse.Pedigree, I set metafounders for all of the individuals for simplicity, so the progenies of the progenies of different metafounders all share a default metafounder, so we don't want to update their anterior probabilities otherwise it would interfere with the information while passing down the pedigree if the alternative allele probability of the default metafounder is not 0.5.

gregorgorjanc:
@RosCraddock thanks for using the warning!

@RosCraddock @XingerTang I think we are nearing to a merge of this PR. I started from the top and tried to check if we covered everything discussed, and it seems we did! Well done @RosCraddock! Am I missing anything left out (squash of the commits I reckon)?

I am only confused about the bit mentioned in https://github.com/AlphaGenes/AlphaPeel/pull/175#discussion_r1815992476 - should we discuss this here or in person and bring this PR to a merge?

src/tinypeel/tinypeel.py
RosCraddock: I can see in Tinyhouse that, by default, all individuals in the pedigree are assigned a metafounder. This is either the default (MF_1) or the main_metafounder defined by the user. The founders' metafounder is later updated to the one they have been allocated per the pedigree input. Thus, to update the anterior probability of the founders, we must use the metafounder and check the individual is a founder with `isFounder()`. Otherwise, we could unintentionally update the anterior probabilities of the progeny with the allele frequency of the main metafounder (i.e the default of 0.5). This would be unnecessary and reduce computational efficiency. However, I don't know how this impacts peeling since the anterior probabilities of the progeny are later informed by the parents. @XingerTang could you clarify your last sentence in the above [comment](https://github.com/AlphaGenes/AlphaPeel/pull/175#discussion_r1745131168)?

XingerTang: @RosCraddock is right that most of the time, the anterior probabilities of the progenies are recalculated and overwritten using the probabilities from the parents during the peeling, so it would only lead to the waste of time and computational power but not affect the result. It will cause a problem if we try to peel down the pedigree but not in the right order of generations, in this case, the anterior probabilities from the default metafounders would be used to calculate the genotype probabilities of all of the individuals and pass down or up. But these are all fixed with the conditioning with `isFounder()`.

gregorgorjanc: OK, I think all is well then.

Before we resolve this conversation and move onwards, I feel that this is an implementation detail that can easily get lost with time and could be a source of hard to figure out bugs. We should make a code comment somewhere! @XingerTang @RosCraddock please work on this comment as part of this PR. Thanks!

RosCraddock: @gregorgorjanc That's a good point! I will add comments in tinyhouse and tinypeel.py to clarify this.

gregorgorjanc: @RosCraddock thanks for doing this at https://github.com/AlphaGenes/tinyhouse/pull/19

gregorgorjanc:
@XingerTang if you have time I would appreciate if you can look into this - I am in grant and manuscript mode just now;)

RosCraddock:
> @RosCraddock silly question and sorry if I overlooked this, but how do you switch as a user between the current default Newton update of allele frequency for whole population (works with one MF only)and our EM-like update of allele frequency based on founder inferred genotype probabilities (with one or more MF)?

@gregorgorjanc It's not a silly question, and you have not overlooked it! I have yet to push any code for estimating the alternative allele frequency with multiple metafounders as I was waiting for this PR to be completed. Then, I can start another PR for estimating allele frequencies. I have yet to implement the user-input side, but I intend to add another command called `-update_alt_allele_prob_after_peeling` (or `-update_alt_allele_prob_for_metafounders`) and update the documentation for multiple metafounders. This command would function with `-est_alt_allele_prob`, `-alt_allele_prob_file`, or using default allele frequencies whether there were one or more metafounders.

gregorgorjanc:
> @gregorgorjanc It's not a silly question, and you have not overlooked it! I have yet to push any code for estimating the alternative allele frequency with multiple metafounders as I was waiting for this PR to be completed. Then, I can start another PR for estimating allele frequencies. I have yet to implement the user-input side, but I intend to add another command called `-update_alt_allele_prob_after_peeling` (or `-update_alt_allele_prob_for_metafounders`) and update the documentation for multiple metafounders. This command would function with `-est_alt_allele_prob`, `-alt_allele_prob_file`, or using default allele frequencies whether there are one or more metafounders.

Gotcha! @RosCraddock Let's talk in the new PR on how to go about doing this from the option point of view. I think we probably want to stick only to  `-est_alt_allele_prob` and `-est_alt_allele_prob_method` or something similar and whether this is done for 1 or more MF is simply a function of the data input (current Newton will not work for multiple MF - would need adaptation).

XingerTang:
> Sorry, it seems I have added a conflict with the submodule tinyhouse when I squashed the commits. Not sure how I can resolve this? @XingerTang @gregorgorjanc
> 
> (edited) I have checked all the files, and all are matching the last commit in tinyhouse/feat_metafounders (6e368443b186164ea889fa01de1a19c1aa5d2409). There were no conflicts before squashing commits, and I did not introduce anything else. So not sure why it is happening.

@RosCraddock I saw you committed a new reference update to the tinyhouse, and it seems up to date for me. Is your problem now resolved?

RosCraddock:
> @RosCraddock I saw you committed a new reference update to the tinyhouse, and it seems up to date for me. Is your problem now resolved?
> 
> Or if your problem is with squashing, could you provide the exact git instructions you used to squash the changes?

@XingerTang I thought that may resolve it, but the conflict still remains. For the squashing, I followed the instructions in this stack overflow: https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together

```
git reset --soft HEAD~17
git commit
git push --force-with-lease
```

RosCraddock:
I think I have corrected my mistake and the conflict that occurred. I essentially reverted the squash to a new branch, squashed some commits I was able to without introducing the conflict, and then merged the new branch into the feat_metafounders branch. I am reluctant to try squashing the commits any further in case I introduce another conflict.
@gregorgorjanc @XingerTang

XingerTang:
@RosCraddock Thank you for your work! It looks good to me. But for the squashing, maybe we need to find a better way to work with it and the submodule reference.

PR #182, opened 2024-11-12
Estimation of alternative allele frequencies with metafounders

RosCraddock:
Quite a long list of updates here!
- New command `-update_alt_allele_prob` for re-estimating the alternative allele frequency on each peeling
- New function updateMafAfterPeeling() to re-estimate the alternative allele frequencies based on the inferred genotype probabilities of the founders (triggered with new command
- Restricted all alternative allele frequencies to be between 0.01 and 0.99 (including user-inputs).
- Updates to documentation for -update_alt_allele_prob and noted the alternative allele frequency range (0.01 to 0.99).
- Updates to functional testing for alternative allele frequency estimation.
- New R script file for metafounder simulation (new code replacing the previous approach).
- New simulation for accuracy testing of metafounders: one pedigree, one true genotype (dosage, phased, unphased), set of obsv genotypes with ~50% missing rate.
- Five additional accuracy tests for metafounders. All only with the single method. I can add more (e.g multi, hybrid), but cautious since the summary table is already quite large.
- Additions to the accuracy summary table printed in the terminal to include metafounder (as separate entry as uses different pedigree simulation).

@gregorgorjanc @XingerTang - when you have time, please let me know if you have any comments/questions/changes.


docs/source/usage.rst
```diff
@@ -113,9 +113,12 @@ Peeling arguments
                             each peeling cycle.
       -est_rec_prob         Flag to re-estimate the recombination rates after
                             each peeling cycle.
-      -est_alt_allele_prob  Flag to re-estimate the alternative allele probabilities for each metafounder after
-                            each peeling cycle.
-      -no_phase_founder    A flag phase a heterozygous allele in one of the
+      -est_alt_allele_prob  Flag to estimate the alternative allele frequencies using
+                            all observed genotypes prior to peeling.
+      -update_alt_allele_prob
```
gregorgorjanc, line 118: @RosCraddock we will probably experiment a bit with this. I like and dislike the estimate and update options we have here. It’s a good approach I think nevertheless. The dislike part is that it’s another option, but it’s ok - it gives us flexibility.

At the moment the “estimation” uses the “Newton” path in the code on observed genotypes anywhere in the pedigree, but only with one metafounder. The “update” uses “EM” path in the code on estimated genotype probabilities in founders, for any number of metafounders.

Maybe we find a way to expand the “Newton” path, so indeed best to keep the options separate.

But, to make it lighter on users should we activate estimation when we set the updating? Hmm, I guess not, because current default is to use 0.5, which would not be activated at all. Ok, I just had to write this up to clear everything in my head;) All good, what we have is very good set of options.

docs/source/usage.rst
```diff
@@ -133,7 +136,7 @@ For hybrid peeling, where a large amount (millions of segregating sites) of sequ
...
+The ``-est_geno_error_prob`` and ``-est_seq_error_prob`` options estimate the genotyping error rate and the sequencing error rate based on miss-match between observed and inferred states. This option is generally not necessary and can increase runtime. ``-est_alt_allele_prob`` estimates the alternative allele frequencies before peeling using all available observed genotypes. This option can be useful if there are a large number of non-genotyped founders. ``-update_alt_allele_prob`` re-estimates the alternative allele frequencies per metafounder after each peeling cycle using the inferred genotype probabilities of the founders. For implementation of metafounders (**without** ``-alt_allele_prob_file``), both ``-est_alt_allele_prob`` and ``-update_alt_allele_prob`` should be used.
```
gregorgorjanc, line 139: The last sentence on providing alternative allele prob file - I don’t fully follow what you mean here. Do you need to say more?

docs/source/usage.rst
```diff
@@ -342,7 +345,7 @@ Example:
...
+|Software| outputs four model parameter files: *.alt_allele_prob.txt*, *.seq_error_prob.txt*, *.geno_error_prob.txt*, *.rec_prob.txt*. These give the alternative allele frequency, sequencing error rates, genotyping error rates and the recombination rates used. In the *.alt_allele_prob.txt*, there is a column per metafounder with an alternative allele frequency for each marker. Here, all values will range from 0.01 to 0.99. The other three files contain a single column with an entry for each marker. By default, |Software| will output *.seq_error_prob.txt*, *.geno_error_prob.txt* and *.rec_prob.txt*. The *.alt_allele_prob.txt* will only be outputted with the argument ``-alt_allele_prob``.
```
gregorgorjanc, line 348: @XingerTang we limit ourselves to 1%-99% to avoid getting “sucked” into 0, where then everything becomes “fixed”. Should we have a bit more lax range? Say, 0.001-0.999 or similar? How to do this well from numerical perspective?

src/tinypeel/Peeling/PeelingUpdates.py
```diff
@@ -111,6 +111,34 @@ def addIndividualToUpdate(d, p, LLp, LLpp):
     return LLp, LLpp
 
 
+def updateMafAfterPeeling(pedigree, peelingInfo):
```
gregorgorjanc, line 114: @RosCraddock @XingerTang the code loops over metafounders and for each metafounder loops over all pedigree members twice - once to calculate and then to assign anterior probabilities. I am thinking out loud if we can speed this up in any way if we flip iteration around - over individuals and then over metafounders, but we would be doing the same amount of work, so it seems we can not.

src/tinypeel/tinypeel.py
```diff
@@ -28,6 +28,14 @@ def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
                     pedigree.AAP[mfx] = np.full(
                         peelingInfo.nLoci, 0.5, dtype=np.float32
                     )
+                else:
+                    AAP = pedigree.AAP[mfx]
+                    for i in range(peelingInfo.nLoci):
```
gregorgorjanc, line 33: @RosCraddock @XingerTang in R, for loops are slow and vectorised work with vectors is much faster. Is it the same in Python? I am just mindful if we have many loci if such bits will add to compute time a lot - not this bit on it’s own, but collectively. We probably want to optimise for speed later if and when that becomes an issue, so just thinking out loud again;)

src/tinypeel/tinypeel.py
```diff
@@ -48,6 +56,10 @@ def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
                         peelingInfo.nLoci, 0.5, dtype=np.float32
                     )
     if args.est_alt_allele_prob:
+        if args.alt_allele_prob_file is not None:
+            warnings.warn(
+                "-est_alt_allele_prob will update the inputted alternative allele frequencies using all available observed genotypes. Therefore, will overwrite any differences between metafounders."
```
gregorgorjanc, line 61: @RosCraddock say that current implementation overwrites differences. We might improve upon this later …

tests/accuracy_tests/sim_for_alphapeel_accu_test/sim_for_metafounder_accu.R
```diff
@@ -0,0 +1,267 @@
+rm(list = ls())
+# install.packages(pkg = "AlphaSimR")
+library(AlphaSimR)
+
+# ---- Two (or more) sub-populations A and B which cross in later generations ----
```
gregorgorjanc, line 5: @RosCraddock great reuse of the simulation code with two sub-populations;)

gregorgorjanc: We probably want to drop phenotypic and genetic values here to make it smaller.

tests/accuracy_tests/sim_for_alphapeel_accu_test/sim_for_metafounder_accu.R
```diff
@@ -0,0 +1,267 @@
...
+                                    fid = pop@father,
+                                    gv1 = pop@gv[, 1],
+                                    pv1 = pop@pheno[, 1],
+                                    gv2 = pop@gv[, 2],
+                                    pv2 = pop@pheno[, 2]))
+  data$haploIBD <- rbind(data$haploIBD,
+                         pullIbdHaplo(pop = pop))
+  data$genoIBS <- rbind(data$genoIBS,
+                        pullSegSiteGeno(pop = pop))
+  listLoci <- seq.int(from = 1, to = sum(pop@nLoci), by = 1)
```
gregorgorjanc, line 59: @RosCraddock do you need this listLoci bit of code? Also, isn't your Genotype the same as genoIBS? I got the later via pullSegSiteGeno() and there is also pullSegSiteHaplo() which will give you haploIBS. That's all you need I think,

tests/accuracy_tests/sim_for_alphapeel_accu_test/sim_for_metafounder_accu.R
```diff
@@ -0,0 +1,267 @@
...
+  marker <- ""
+  i <- 1
+  for (i in 1:length(listLoci)){
```
gregorgorjanc, line 62: Here is an example where doing just paste(“1_”, 1:pop@nLoci, sep=“0”) would be a simpler and faster vectorised way of doing this, but you don’t need it at all - just use pullSegSiteHaplo(‘)

tests/accuracy_tests/sim_for_alphapeel_accu_test/sim_for_metafounder_accu.R
gregorgorjanc: As in, no need to collect them

tests/accuracy_tests/sim_for_alphapeel_accu_test/sim_for_metafounder_accu.R
```diff
@@ -0,0 +1,267 @@
...
+data$pedigree$gv <- c(selIndex(Y = as.matrix(data$pedigree[, c("gv1", "gv2")]), scale = TRUE, b = selIndexWeights))
+data$pedigree$pv <- c(selIndex(Y = as.matrix(data$pedigree[, c("pv1", "pv2")]), scale = TRUE, b = selIndexWeights))
+
+data$pedigree$generationPlotShift <- data$pedigree$generation +
+  c(-0.25, +0.25, 0)[data$pedigree$population]
+
+# Get pedigree and assign metafounders
+pedigree <- data.frame(data$pedigree$id, data$pedigree$mid, data$pedigree$fid, data$pedigree$population, data$pedigree$generation)
+colnames(pedigree) <- c("id", "mid", "fid", "population", "generation")
+pedigree <- pedigree[c(401:1400),]
```
gregorgorjanc, line 144: We have hardcoded numbers here but nGen comes into this script externally - will that clash with these hardcoded numbers here?

gregorgorjanc (commented):
Very good work @RosCraddock. I left some minor comments.

docs/source/usage.rst
RosCraddock: I will rephrase this. Here, I wanted to highlight that `-est_alt_allele_prob` estimates one alternative allele frequency for the base population using all available genotype data (so it ignores any genetic grouping). Thus, to utilise the metafounders/genetic grouping in the peeling, `-update_alt_allele_prob` is required as well.

src/tinypeel/Peeling/PeelingUpdates.py
RosCraddock: Yes, I wondered the same when writing this!

src/tinypeel/tinypeel.py
RosCraddock: I think it is the same in Python!

src/tinypeel/tinypeel.py
RosCraddock: This is only true if a user uses both `—alt_allele_prob_file` and `—est_alt_allele_prob`. For now, I think I will rewrite this warning to recommend using `—update_alt_allele_prob` instead.

tests/accuracy_tests/sim_for_alphapeel_accu_test/sim_for_metafounder_accu.R
RosCraddock: Yes, sorry - I forgot to remove the redundant code!

docs/source/usage.rst
RosCraddock: I think we could enforce a more laxed range. I did a quick test with inputted alternative allele frequencies and the estimation via the Newton method, and it seems to be fine. But it would be good to know what @XingerTang thinks!

gregorgorjanc: @RosCraddock I was asking this because some disease allele frequencies in a base population might be very low, much lower than 0.01, but we do want to avoid the "0 trap", which can occur during iteration/peeling cycles. The question now is what limits should we put. @XingerTang I would appreciate your math's input here;)

gregorgorjanc: @RosCraddock one way to check this is to see in your real datasets you work with - if you get allele frequency that is 0.01 for any of the metafounders then we know that we have likely hit this imposed range-limit. That will give us a good test for how to manage these range-limits.

gregorgorjanc (approved):
Looks good @RosCraddock

docs/source/usage.rst
gregorgorjanc: @RosCraddock Have you had a chance to look into this?

gregorgorjanc (approved):
LGTM - we can deal with 0.01 - 0.99 bit later - will open an issue

gregorgorjanc:
@RosCraddock can you please squash the commits into one and give it commit message "Estimation of alternative allele frequencies with metafounders". Then we merge this in. Thanks!

RosCraddock:
I cannot squash the commits further (from 8 to 3 now) due to a complex conflict with a submodule update from the previous merge. 

I was not expecting the error in the checks (I only squashed commits, and very carefully too). Looks like the issue is in the pip installation. The wheel file is saved as alphapeel-1.1.3 etc, but for installation we search for Alphapeel-1.1.3 etc. Thus, fails as the "file does not exist". @XingerTang. Can you point me to the code for this so I can resolve this?

RosCraddock:
I found the cause for the error above - on 25th Feb the setuptools package was updated for all wheel file naming to follow binary distribution specification (i.e all lower case). It states that this is a trial (so they may revert?). I will set up an issue to resolve this - this will be relevant for other branches too.

gregorgorjanc:
@RosCraddock thanks for debugging this one. I am merging this in! Well done!

PR #191, opened 2025-04-23
Commenting of functions

RosCraddock:
Short description for each function in tinypeel and PeelingIO.
Will continue for the other files!

### DIFF
--- a/src/tinypeel/Peeling/PeelingIO.py
+++ b/src/tinypeel/Peeling/PeelingIO.py
@@ -4,6 +4,9 @@ from ..tinyhouse import InputOutput
 
 
 def readInSeg(pedigree, fileName, start=None, stop=None):
+    """
+    Reads in a segregation file and returns a 3D numpy array of segregation probabilities.
+    """
     print("Reading in seg file:", fileName)
     if start is None:
         start = 0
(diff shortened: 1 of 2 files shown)


XingerTang:
@RosCraddock 

All the docstrings look good to me! I also did similar work with the `src/Peeling.py` file, which you can find at [here](https://github.com/XingerTang/AlphaPeel/blob/ea4615c7f9679bcf0d3b34edfbe231239fbec330/src/tinypeel/Peeling/Peeling.py). There are some changes in the code, but they are all made inside the functions. All the input and output variables are the same, so the docstrings are not affected.

One standard format for the docstring is the [Sphinx format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html). If you want to use the format, you can try the VSCode extension [autoDocstring](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) and set the "autoDocstring.docstringFormat": "sphinx" setting, under File > Preferences > Settings, to automatically generate templates of docstrings for each function.

One of the advantages of using such a format is that, as our documentation is also generated with Sphinx, we can directly reference the function with its details in our documentation, such as [this section](https://xingertang-alphapeel.readthedocs.io/en/latest/usage.html#code-explanation).

Issue #184, opened 2025-02-19
Change -sex_chrom to -x_chr


AprilYUZhang:

XingerTang:
@AprilYUZhang May I ask how you would handle the tinyhouse reference by the other Alpha packages that may use the ``sex_chrom`` flag

AprilYUZhang:
I have applied for a [pull request](https://github.com/AlphaGenes/tinyhouse/pull/24).

gregorgorjanc:
Since sex chromosome is experimental and not really working in AlphaImpute2, then I don't think this is an issue and we can break backwards compatibility with this one. Anyway, sex_chromosome isn't a good name anyway, because there is X and Y chromosome behind the "sex chromosome", albeit many use the "sex chromosome" as synonym for X in a sense that Y might be easy to deal with.

Issue #112, opened 2023-10-09
What does the peeling option -nophasefounders do?


gregorgorjanc:
As part of clarifying the documentation I am trying to understand what does `-nophasefounders` option do so we can write better docs!?

The docs say `A flag phase a heterozygous allele in one of the founders (if such an allele can be found).`

It's used here https://github.com/AlphaGenes/AlphaPeel/blob/1c56d89e49b78b23c6ef6b3651164448697097f7/src/tinypeel/Peeling/PeelingUpdates.py#L131

https://github.com/AlphaGenes/AlphaPeel/blob/1c56d89e49b78b23c6ef6b3651164448697097f7/src/tinypeel/tinypeel.py#L250C65-L250C77

https://github.com/AlphaGenes/AlphaPeel/blob/1c56d89e49b78b23c6ef6b3651164448697097f7/src/tinypeel/Peeling/PeelingInfo.py#L19C55-L19C67

https://github.com/AlphaGenes/AlphaPeel/blob/1c56d89e49b78b23c6ef6b3651164448697097f7/src/tinypeel/Peeling/PeelingInfo.py#L57

apart from definitions

https://github.com/AlphaGenes/AlphaPeel/blob/1c56d89e49b78b23c6ef6b3651164448697097f7/src/tinypeel/tinypeel.py#L222

XingerTang:
@gregorgorjanc I read the codes

If the `nophasefounders` flag is `True`, then the program would do the following:

If both parents of the genotyped individual are not genotyped (so the individual itself is a genotyped founder), then the penetrance for genotypes `aa`, `aA`, `Aa`, `AA` at the closest heterozygous locus to the midpoint locus of the whole genotype of the individual would be set to `geno_error_prob / 3`, `geno_error_prob / 3`, `1 - geno_error_prob`, `geno_error_prob / 3` respectively. 
And, this would be done at the initialization stage as well as each time the penetrance for each individual is updated.

So the use of the flag is to phase the genotype at one unphased heterozygous locus of a genotyped individual without genotyped parents to `Aa` instead of equal probabilities of `aA` and `Aa`.  

One thing that I am curious to is why the program only phases one particular heterozygous locus that is closest to the midpoint, but instead of all the heterozygous loci of the individual. But otherwise, I think the flag makes sense,  the only problem is it has a bad name.

RosCraddock:
[...]

PR #192, opened 2025-05-09
Sex chrom

AprilYUZhang:


docs/source/usage.rst
```diff
@@ -122,26 +122,27 @@ Peeling arguments
                             for each metafounder after each peeling cycle.
       -no_phase_founder     A flag phase a heterozygous allele in one of the
                             founders (if such an allele can be found).
-      -sex_chrom            A flag to indicate that input data is for a sex chromosome. Sex needs to
-                            be given in the pedigree file. This is currently an
-                            experimental option.
-
+      -sex_chrom            A flag to indicate that input data is for a sex chromosome.
```
gregorgorjanc, line 125: Let's move to `x_chrom` as mentioned in https://github.com/AlphaGenes/AlphaPeel/issues/184

docs/source/usage.rst
```diff
@@ -122,26 +122,27 @@ Peeling arguments
...
 
 For a pedigree with multiple metafounders, the user has three options: (1) use ``-update_alt_allele_prob`` only, (2) use ``est_alt_allele_prob`` and ``-update_alt_allele_prob``, or (3) input estimates of the alternative allele frequencies for each metafounder via the ``-alt_allele_prob_file`` with or without ``-update_alt_allele_prob``.
 
+When ``-sex_chrom`` is used, there are two requires for input file: (1) The pedigree file must include sex information in the fourth column (0 indicates male, 1 indicates female). See the Pedigree File Format section for an example. (2) male's genotype must be coded as {0,1}. See the Genotype file Format section for an example. Note: This option currently only considers the regions of the X chromosome excluding the pseudoautosomal regions (PARs)
```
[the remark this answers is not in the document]
gregorgorjanc: When ``-x_chrom`` is used, there are two requirements for input files: (1) The pedigree file must include sex information in the fourth column (0 indicates male and 1 indicates female). See the Pedigree File Format section for an example. (2) Male's genotype must be coded as {0,1}. See the Genotype file Format section for an example. Note: This option currently only considers the regions of the X chromosome excluding the pseudo-autosomal regions (PARs).

docs/source/usage.rst
```diff
@@ -185,6 +186,16 @@ or
   id3 id1 id2
   id4 id1 id2
 
+If work on sex chromosome. 0 refer to male, and 1 refer to female.
```
gregorgorjanc, line 189: When working with X chromosome use this format - add sex information in the fourth column (0 indicates male and 1 indicates female):

docs/source/usage.rst
```diff
@@ -199,6 +210,14 @@ Example:
   id3 2 0 2 0 
   id4 0 2 1 0
 
+If work on sex chromosome.
```
gregorgorjanc, line 213: When working with X chromosome, male's genotype must be coded as {0,1}:

src/tinypeel/Peeling/Peeling.py
```diff
@@ -16,7 +16,7 @@
 def peel(family, operation, peelingInfo, singleLocusMode):
     isSexChrom = peelingInfo.isSexChrom
 
-    e = 0.000001
+    e = 0
```
gregorgorjanc, line 19: Revert back please.

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -198,31 +198,36 @@ def writeGenoProbs(pedigree, genoProbFunc, outputFile):
                     )
 
 
-def writeDosages(pedigree, genoProbFunc, isSexChrom, outputFile):
+def writeDosages(pedigree, genoProbFunc, isSexChrom, outputFile):  # remove isSexChrom,
```
gregorgorjanc, line 201: remove the comment

src/tinypeel/Peeling/PeelingInfo.py
```diff
@@ -25,14 +25,25 @@ def createPeelingInfo(pedigree, args, createSeg=True, phaseFounder=False):
     # Information about the peeling positions are handled elsewhere.
     peelingInfo.positions = None
 
+    mutation_rate = args.mutation_rate
     # Generate the segregation tensors.
-    peelingInfo.segregationTensor = ProbMath.generateSegregation(e=1e-06)
-    peelingInfo.segregationTensor_norm = ProbMath.generateSegregation(
-        e=1e-06, partial=True
+    peelingInfo.segregationTensor = ProbMath.generateSegregation_mu(mu=mutation_rate)
```
gregorgorjanc, line 30: ProbMath.generateSegregation_mu --> ProbMath.generateSegregation

src/tinypeel/Peeling/PeelingUpdates.py
```diff
@@ -195,10 +195,15 @@ def updatePenetrance(pedigree, peelingInfo, args):
         ):
             loci = PeelingInfo.getHetMidpoint(ind.genotypes)
             if loci is not None:
-                e = peelingInfo.genoError[loci]
-                peelingInfo.penetrance[ind.idn, :, loci] = np.array(
-                    [e / 3, e / 3, 1 - e, e / 3], dtype=np.float32
-                )
+                error = peelingInfo.genoError[loci]
+                if sexChromFlag:
```
gregorgorjanc, line 199: rename sexChromFlag

tests/accuracy_tests/run_accu_test.py
```diff
@@ -39,6 +39,7 @@ def generate_command(
     alt_allele_prob_file,
     update_alt_allele_prob,
     metafounder,
+    sex_chrom,
```
gregorgorjanc, line 42: x_chrom or x_chr (check the issue)

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -42,7 +42,7 @@ def readInSeg(pedigree, fileName, start=None, stop=None):
                 )
 
             if idx not in pedigree.individuals:
-                print(f"Individual {idx} not found in pedigree. Individual ignored.")
+                print(f"Individual {idx} is found not in pedigree. Individual ignored.")
```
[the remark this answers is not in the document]
AprilYUZhang: I changed this to `print(f"Individual {idx} is not found in pedigree. Individual ignored.")`
because the original one would report error
```
(.venv) (base) s2587593@S31-4GDY27LX AlphaPeel % pre-commit run                           
black....................................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

src/tinypeel/Peeling/PeelingIO.py:45:42: E713 test for membership should be 'not in'
```

gregorgorjanc (approved):
@AprilYUZhang this all looks good to me. Well done! Let's wait now on the tinyhouse PR https://github.com/AlphaGenes/tinyhouse/pull/24 to get merged in so that we can update which commit from that repo you need to include in this PR for AlphaPeel.

gregorgorjanc:
@AprilYUZhang add also a line to CHANGELOG like you did for tinyhouse.

src/tinypeel/Peeling/PeelingUpdates.py
```diff
@@ -195,10 +195,15 @@ def updatePenetrance(pedigree, peelingInfo, args):
         ):
             loci = PeelingInfo.getHetMidpoint(ind.genotypes)
             if loci is not None:
-                e = peelingInfo.genoError[loci]
-                peelingInfo.penetrance[ind.idn, :, loci] = np.array(
-                    [e / 3, e / 3, 1 - e, e / 3], dtype=np.float32
-                )
+                error = peelingInfo.genoError[loci]
```
XingerTang, line 198: You didn't edit line 193; I just noticed that line 193 is using ``(not InputOutput.args.no_phase_founder)``, while previously we have already used the equivalent ``phaseFounder`` flag

src/tinypeel/Peeling/PeelingInfo.py
```diff
@@ -21,18 +21,29 @@ def createPeelingInfo(pedigree, args, createSeg=True, phaseFounder=False):
...
+    )
+    peelingInfo.segregationTensorXY_norm = ProbMath.generateSegregationXYChrom(
+        mu=mutation_rate, partial=True
+    )
+    peelingInfo.segregationTensorXX = ProbMath.generateSegregationXXChrom(
+        mu=mutation_rate
+    )
+    peelingInfo.segregationTensorXX_norm = ProbMath.generateSegregationXXChrom(
+        mu=mutation_rate, partial=True
+    )
```
XingerTang, line 46: These are only needed when ``isXChr`` is ``True``, maybe add a conditional for the sake of performance.

src/tinypeel/Peeling/PeelingInfo.py
```diff
@@ -69,15 +80,20 @@ def createPeelingInfo(pedigree, args, createSeg=True, phaseFounder=False):
         if ind.isGenotypedFounder() and phaseFounder and ind.genotypes is not None:
             loci = getHetMidpoint(ind.genotypes)
             if loci is not None:
-                e = args.geno_error_prob
-                peelingInfo.penetrance[ind.idn, :, loci] = np.array(
-                    [e / 3, e / 3, 1 - e, e / 3], dtype=np.float32
-                )
+                error = args.geno_error_prob
+                if args.x_chr and ind.sex == 0:
```
XingerTang, line 84: Maybe we can use ``peelingInfo.isXChr`` consistently, instead of some being ``args.x_chr``

XingerTang: same applied to the later occurrence in line 94

tests/functional_tests/run_func_test.py
```diff
@@ -76,6 +76,53 @@ def delete_columns(two_d_list, col_del):
...
+    """
+    Compare the output file with the true file
+    the error tolerance is mismatch genotype <= total_error
+    """
+    outputs = read_geno_hap(output)
+    trues = read_geno_hap(true)  
+
+    trues_id = sorted(trues.keys())
+    outputs_id = sorted(outputs.keys())
+    # check the number of lines
```
XingerTang, line 105: The comment does not match what the check is doing. I suppose we are checking whether the individual ids are matched.

src/tinypeel/Peeling/PeelingInfo.py
XingerTang, line 46: Same as the comment in the tinyhouse pull request AlphaGenes/tinyhouse#24, maybe we don't need a duplicate segregation tensor.

gregorgorjanc: Agreed!

src/tinypeel/Peeling/PeelingInfo.py
```diff
@@ -69,15 +80,20 @@ def createPeelingInfo(pedigree, args, createSeg=True, phaseFounder=False):
...
+                    peelingInfo.penetrance[ind.idn, :, loci] = np.array(
+                        [error / 3, 1 - error, error / 3, error / 3], dtype=np.float32
+                    )
+                else:
+                    peelingInfo.penetrance[ind.idn, :, loci] = np.array(
+                        [error / 3, error / 3, 1 - error, error / 3], dtype=np.float32
+                    )
```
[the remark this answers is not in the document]
XingerTang: The original chosen vector is related to the discussion in #112, that the original choice of the ``[error / 3, error / 3, 1 - error, error / 3]`` as ``Aa`` is arbitrary, and can be modified as the other way round with ``aA`` without making any actual changes in its use (that to have a phased state one way or another). 

I propose to add the ``0``s, because @AprilYUZhang is only using the probabilities associated with the ``aA`` and ``AA`` states among the four for the males sex chromosome, which leads to my suggestion that maybe ruling out the other probabilities completely might make more sense.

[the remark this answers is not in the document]
AprilYUZhang: I wrote it for keeping consistent with the original code. This change would not affect too much, I can follow you guys. If add 0 my suggestion is `[error / 2, 1- error, error/2, 0] ` because ind.sex == 0 means male.

src/tinypeel/Peeling/PeelingUpdates.py
AprilYUZhang: They are different functions, the input parameters are different, and Consistency requires greater changes。

src/tinypeel/Peeling/PeelingInfo.py
AprilYUZhang: [09a4773](https://github.com/AlphaGenes/AlphaPeel/pull/192/commits/09a4773a7dd7d8132d017dda435807c7f5d8ebc2)
Add!

src/tinypeel/Peeling/PeelingInfo.py
AprilYUZhang: [45ad327](https://github.com/AlphaGenes/AlphaPeel/pull/192/commits/45ad327fef58df22f107d6e19c0fa83f21540f26)
Here it is

src/tinypeel/Peeling/PeelingInfo.py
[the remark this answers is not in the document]
XingerTang: I was suggesting that way because in the 
https://github.com/AlphaGenes/AlphaPeel/pull/192/files#diff-698ebee25d189e41fb90d5984a5ff406ab594db871ec3adce4b57cc23ac58c43R205

You are using 
```
np.array([0, 1, 0, 1])
```
to filter out the genotype probabilities output for the male sex chromosome, and here you are also handling the male sex chromosome, I am expecting you to use a similar choice.


Your suggestion of ``[error / 2, 1- error, error/2, 0] ``corresponds to the highest genotype probability in ``aA``, but errors fall into ``aa`` and ``Aa``. I don't quite understand the distribution here.

src/tinypeel/Peeling/PeelingUpdates.py
XingerTang: In https://github.com/AlphaGenes/AlphaPeel/pull/192/files#diff-1e6075458414cbc541b5daa7ceae90d968fb8c2d7158b0596cca579685dc6789L664, we have set 
```
phaseFounder=(not args.no_phase_founder)
```

AprilYUZhang: here it is [5f1cb7b](https://github.com/AlphaGenes/AlphaPeel/pull/192/commits/5f1cb7b2eb31a3988fc736d946683a250ba3a3dc)

src/tinypeel/Peeling/PeelingInfo.py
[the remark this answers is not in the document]
gregorgorjanc: > ind.sex == 0 means male.

@AprilYUZhang when you revise, add a comment that sex = 0 is male to the code so it's easier to read it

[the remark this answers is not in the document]
AprilYUZhang: np.array([0, 1, 0, 1]) is not penetrance, for dosage, we only care about the allele from the mom,  for example, aa, aA, Aa, AA, we consider the latter one. so it is the [0,1,0,1].
[error / 2, 1- error, error/2, 0]  is generated from the perspective of penetrance, we can only observe the  0,1, so the forth is not possible.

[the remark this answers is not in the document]
XingerTang: I understand what brings you to your choice, but I just rethought the whole part of the code, and found that probably the sex chromosome should not be part of this at all. 

This part is meant to phase the undetermined heterozygous state to one of the two possible states for the ease of later calculation, while the concept of the heterozygous state itself doesn't apply to the sex chromosome. The best thing we can get from here is that we assign the penetrance such that the genotype probabilities of the sex chromosomes are not affected by this reassignment. So I would suggest that we modify the conditions in line 80 from 
```
if ind.isGenotypedFounder() and phaseFounder and ind.genotypes is not None:
```
to 
```
if ind.isGenotypedFounder() and phaseFounder and ind.genotypes is not None and not peelingInfo.isXChr:
```
or equivalent, to exclude the sex chromosome from this piece of code that should only be applied on the heterozygous state.

Also, I would suggest that we can modify the ``np.array([0, 1, 0, 1])`` of the sex chromosome dosage output to ``np.array([0, 1, 0, 0])``. It won't change the result, as the sex chromosome won't have the state ``AA``. But, it would make more sense when reading it and possibly saves unnecessary computation.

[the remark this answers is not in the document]
AprilYUZhang: XX also need to consider the heterozygous state in females, but I agree with you to change the conditions from
`if peelingInfo.isXChr and ind.sex == 0: # sex = 0 is male
                    peelingInfo.penetrance[ind.idn, :, loci] = np.array(
                        [error / 3, 1 - error, error / 3, error / 3], dtype=np.float32
                    )
                else:
                    peelingInfo.penetrance[ind.idn, :, loci] = np.array(
                        [error / 3, error / 3, 1 - error, error / 3], dtype=np.float32
                    )`
to 
`if (not peelingInfo.isXChr) or (ind.sex != 0): # sex = 0 is male':
                    peelingInfo.penetrance[ind.idn, :, loci] = np.array(
                        [error / 3, error / 3, 1 - error, error / 3], dtype=np.float32
                    )`
From the perspective of efficiency, we can use this condition earlier.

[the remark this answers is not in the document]
AprilYUZhang: Regarding Dosage, np.array([0, 1, 0, 1]) is explainable if we calculate dosage from genotype. The genotype of the fourth element should not have a value in theory, but because this is alphapeel,  there are always some very small values, compared to 0.
What I said above doesn't mean I don't agree with you. Both are ok for me.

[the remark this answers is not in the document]
XingerTang: Your condition modification looks great. 
Regarding the dosage, I have discussed with @gregorgorjanc last week, and we both prefer the ``np.array([0, 1, 0, 0])``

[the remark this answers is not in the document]
AprilYUZhang: [improve code efficiency](https://github.com/AlphaGenes/AlphaPeel/pull/192/commits/425e3b543025a2e744efb4543dd5772674f76975)
Here is it.

src/tinypeel/Peeling/PeelingUpdates.py
```diff
@@ -195,10 +195,15 @@ def updatePenetrance(pedigree, peelingInfo, args):
...
+                if XChrMaleFlag:
+                    peelingInfo.penetrance[ind.idn, :, loci] = np.array(
+                        [error / 3, 1 - error, error / 3, error / 3], dtype=np.float32
+                    )
+                else:
+                    peelingInfo.penetrance[ind.idn, :, loci] = np.array(
+                        [error / 3, error / 3, 1 - error, error / 3], dtype=np.float32
+                    )
```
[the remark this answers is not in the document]
XingerTang: All looks good to me. Could you go to the tinyhouse pull request and then have a look at my recent comments?

XingerTang:
@AprilYUZhang could you update the tinyhouse reference and see if the tests pass

XingerTang:
@AprilYUZhang Thanks! Could you trigger the pipeline here so that the GitHub runner runs your test?

PR #198, opened 2025-11-05
X chrom functionality

AprilYUZhang:


XingerTang:
@AprilYUZhang 
There are too many changes that the web editor does not support the conflict resolution here. For me to do that, I probably need to first fork your repository and then submit a pull request to your branch on your repository, which, if you later accept, would be able to add here. This is quite complicated. Also, since I won't be able to access the conflicts directly, I won't be able to merge them here. I can only provide change suggestions. So, I think it is still better for you to resolve the conflicts. 
Or another way is that I create a new branch from the earlier commit of this repository, and then you create a pull request to that branch. In this way, you can first push to that branch and then leave the conflict resolution for later.

Issue #199, opened 2025-11-09
x_chr example showing the needed improvements


AprilYUZhang:
An example so users can try running this functionality and they can also explore the outputs

XingerTang:
> We need an example for the autosomal chr and for the X chr.

gregorgorjanc:
As part of checking the current X chr implementation in https://github.com/AlphaGenes/AlphaPeel/pull/215 I noticed that it doesn't work as it should. Here is a more extended, but still simple example introduced in that PR with comments showing what we have to do.

```

AlphaPeel -geno_file simple_genotype.txt \
          -ped_file simple_pedigree.txt \
          -x_chr \
          -geno -hap -geno_prob -phased_geno_prob -seg_prob \
          -out_file simple_output

Assumed/true haplotypes used to generate the genotypes below
A - - - - - - -
A 1 1 1 1 0 0 0
B 1 1 1 1 1 1 1
B 1 1 1 1 1 1 1
C - - - - - - -
C 1 1 1 1 1 1 1
D 1 1 1 1 0 0 0
D 1 1 1 1 1 1 1
E - - - - - - -
E 1 1 1 1 0 0 1
F 1 1 1 1 1 1 1
F 1 1 1 1 0 0 0
G - - - - - - -
G 1 1 1 1 1 1 1
H - - - - - - -
H 0 0 0 0 0 0 0
I 0 0 0 0 0 0 0
I 0 0 0 0 0 0 0
J 1 1 1 1 1 1 1
J 1 1 1 1 1 1 1

simple_output.hap_0.5.txt looks correct (mostly?)!
A 9 9 9 9 9 9 9
A 1 1 1 1 0 0 0
B 1 1 1 1 1 1 1
B 1 1 1 1 1 1 1
C 9 9 9 9 9 9 9
C 1 1 1 1 1 1 1
D 1 1 1 1 0 0 0
D 1 1 1 1 1 1 1
E 9 9 9 9 9 9 9
E 1 1 1 1 0 0 1
F 1 1 1 1 1 1 1
F 1 1 1 1 0 9 0 --> TODO: Why is there the 9 here?
G 9 9 9 9 9 9 9
G 1 1 1 1 1 1 1
H 9 9 9 9 9 9 9
H 0 0 0 0 0 0 0
I 0 0 0 0 0 0 0
I 0 0 0 0 0 0 0
J 1 1 1 1 1 1 1
J 1 1 1 1 1 1 1

[...]

Issue #211, opened 2026-01-29
Mismatch between haplotype output and genotype output


XingerTang:
This happened once for AlphaImpute2 (https://github.com/AlphaGenes/AlphaImpute2/issues/51), but it is now fixed. 

With input genotype:

```
1 1 1 1 1 1
2 0 0 0 0 0
3 1 1 9 9 0
```

Input pedigree:

```
1 0 0
2 0 0
3 1 2
```

Command:

```
AlphaPeel -geno_file test_genotypes.txt -ped_file test_pedigree.txt -method single -geno -hap -out_file simple_test
```

The output called genotype is:

```
1 1 1 1 1 1
2 0 0 0 0 0
3 1 1 1 1 0
```

While the output called haplotype is:

```
1 1 1 1 9 0
1 1 1 0 9 0
2 0 0 0 0 0
2 0 0 0 0 0
3 1 1 9 9 0
3 0 0 0 0 0
```

The haplotypes of the first two loci of individual 1 are incorrect.

PR #215, opened 2026-02-05
Added an x_chr example

gregorgorjanc:
I have added a simple x_chr example in line with #199. It shows one issue with dosages (they are wrong for males, unless I have wrong tinyhouse? no, I pulled the most recent and still get the same results!) and one issue in the haplotype for one individual.

See the file `examples/simple_example_x/simple_true.txt` for the true X chromosome haplotypes and genotypes. You will have to also look at `simple_pedigree.txt` in the directory to make sense of the example.

I propose the following. @AprilYUZhang and @XingerTang can you please have a look at my changes ~~(I only formatted the docs)~~ and if the example makes sense for X chromosome inheritance (with recombination here) and we get this merged in. Then we go and address the problem with the dosage and the odd 9 in haplotype in a separate issue. I suggest we leave the `simple_true.txt` in the directory the way it is and then we polish it in the next issue.


XingerTang (commented):
All looks good to me!

XingerTang:
Hi @gregorgorjanc , I think what you obtained from the called haplotypes, phased genotype probabilities, and (unphased) genotype probabilities are all expected. 

The ``9`` of the individual ``F`` in the output haplotype is caused by the call threshold set as ``0.5``, since none of its four phased genotype states has a probability over ``0.5``, it is set as unknown. You may find that for the same individual and same locus, the genotype output is correct and gives a confident ``1``, which should suggest what the haplotype output should be at the same locus. This is determined by the output process and is what we are going to solve for Issue https://github.com/AlphaGenes/AlphaPeel/issues/211

The phased genotype probabilities and (unphased) genotype probabilities outputs are expected because we chose to use two states of a pair of regular chromosomes to represent one state of the X chromosome. It is assumed to be read in a way that sums up the first and third rows and then compares the sum of the second and fourth rows for the phased genotype probabilities. While for the unphased, it simply mixed the middle two rows of the phased genotype probabilities together, so the same logic applies. We can create special output functions for the X chromosome for the genotype probabilities in the future to make it easier to interpret.

The dosage does require further normalisation to make sense. For X chromosomes, the dosage is essentially the second row of the phased genotype probabilities. Because in the case of the X chromosome, we use the probability of two states to represent one state, its probability is split up into two parts. So the actual dosage should be two times the output.

For the interpretation of the segregation probabilities of X chromosomes, we can look at the first two rows and get insight into which haplotype of the mother is inherited.

gregorgorjanc:
@XingerTang thanks for looking into this. I have met with @AprilYUZhang today and it seems to me that if we change the penetrance function, then the phased genotype probs will change to just `aa` vs `AA` state and no probability for the `aA` and `Aa` states [speaking for for the males here] (we might need to review the transmission function too?) and theneverything downstream should work out, hopefully. @AprilYUZhang will share her plan on how the change might look like and we will work with her on implementation.

XingerTang:
Should we also add a section in the documentation for this example?

gregorgorjanc:
> Should we also add a section in the documentation for this example?

Yes, but I did this already in https://github.com/AlphaGenes/AlphaPeel/pull/219, though I think you mean something more - can you elaborate? Perhaps expose that there is/will be X chr example?

XingerTang:
> > Should we also add a section in the documentation for this example?
> 
> Yes, but I did this already in #219, though I think you mean something more - can you elaborate? Perhaps expose that there is/will be X chr example?

We have the description and illustration for the plain simple example in the Getting Started section (https://alphapeel.readthedocs.io/en/latest/get_started.html#an-example). I'm expecting maybe we can do something similar for the X chr example.

gregorgorjanc:
We probably want to keep the Getting started section simple, so autosomal chr is fine there. 

We should think of somehow promoting these tiny examples more, particularly once we add X, Y, and mtDNA cases! Can you think of the best way for this @XingerTang?

Issue #230, opened 2026-03-12
Improve technical code instruction of the contribution guide


XingerTang:
## Goal
Improve the technical code instruction of the contribution guide for a clearer and more accurate guide.

## Proposed approach
- [x] Update instructions that do not fit the actual working scene (submodule update)
- [x] Update existing instructions and add new instructions based on the newly added workflow
  - [x] Rebase
  - [x] Squash
  - [x] Test tinyhouse during development
  - [x] Tagging for new release
- [x] Add useful instructions from https://github.com/HighlanderLab/RcppTskit/pull/119/changes#diff-9e530a148d1cbfcb4690a1c1f9aa295b6bd1728bc12afe3c762f56b66f269f47R34
  - [x] ``git`` instructions
- [x] Modify workflow according to discussion in AlphaGenes/AlphaPeel#228
- [x] Add other useful instructions

Others: 
- [x] Add changelog-related section
- [x] Comment the remaining TODOs in documentation (``get_started.rst``)

## Scope
Mainly ``docs/source/usage.rst``, may have minor modifications to other files of documentation 

## Done when
PR open by Mar 18th

Issue #204, opened 2025-12-19
Report version via -version


gregorgorjanc:
Let's implement reporting version and GitHub commit via `-version` flag. Currently I get:

```
❯ AlphaPeel -v
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/bin/AlphaPeel", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/lib/python3.14/site-packages/tinypeel/tinypeel.py", line 242, in main
    args = getArgs()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/lib/python3.14/site-packages/tinypeel/tinypeel.py", line 235, in getArgs
    return InputOutput.parseArgs("AlphaPeel", parser)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/lib/python3.14/site-packages/tinypeel/tinyhouse/InputOutput.py", line 134, in parseArgs
    args = rawParseArgs(program, parser, no_args = no_args)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/lib/python3.14/site-packages/tinypeel/tinyhouse/InputOutput.py", line 168, in rawParseArgs
    with open(args[0]) as f:
         ~~~~^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '-v'
❯ AlphaPeel -version
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/bin/AlphaPeel", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/lib/python3.14/site-packages/tinypeel/tinypeel.py", line 242, in main
    args = getArgs()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/alphasuite/lib/python3.14/site-packages/tinypeel/tinypeel.py", line 235, in getArgs
    return InputOutput.parseArgs("AlphaPeel", parser)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
[...]

PR #237, opened 2026-03-26
Improve technical instruction and add -version flag

XingerTang:
## Related Issue
Closes AlphaGenes/AlphaPeel#230
Closes AlphaGenes/AlphaPeel#204

## What changed
- Update technical instruction, see AlphaGenes/AlphaPeel#230
- Add ``-version`` flag for AlphaGenes/AlphaPeel#204
- Add some links for a quicker start for new users

## Why this change
- Clearer instructions for developers
- Improve usability for users

## Submodule changes
No change made for this PR, but update the reference to the existing most recent commit on the ``devel`` branch of https://github.com/AlphaGenes/tinyhouse (34da21f)

## Notes / Risks
- All the tests passed

### DIFF
--- a/src/tinypeel/tinypeel.py
+++ b/src/tinypeel/tinypeel.py
@@ -16,6 +16,10 @@ import concurrent.futures
 from itertools import repeat
 import argparse
 
+from .Peeling import version
+
+version_version = version.version
+
 
 def runPeelingCycles(pedigree, peelingInfo, args, singleLocusMode=False):
     """Sets up and runs each of the peeling cycles (default is 5).
@@ -825,6 +829,21 @@ def getArgs():
         help="A segregation file for hybrid peeling.",
     )
     # singleLocus_parser.add_argument('-blocksize',default=100, required=False, type=int, help='The number of markers to impute at once. This changes the memory requirements of the program.')
+
+    # special handle for version argument to allow it to be called with just -version
+    parser.add_argument(
+        "-version",
+        default=None,
+        action="version",
+        version="%(prog)s " + version_version,
+        help="Show program's version number and exit.",
+    )
+
+    args = sys.argv[1:]
+
+    if "-version" in args:
+        parser.parse_args(args)
+        sys.exit(0)
 
     return InputOutput.parseArgs("AlphaPeel", parser)
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,16 @@ Whalen, A, Ros-Freixedes, R, Wilson, DL, Gorjanc, G, Hickey, JM. (2018). Hybrid 
 
 See https://alphapeel.readthedocs.io/en/latest
 
+Quick links:
+
+- [Installation](https://alphapeel.readthedocs.io/en/latest/get_started.html#install-software)
+
+- [Simple example](https://alphapeel.readthedocs.io/en/latest/get_started.html#an-example)
+
+- [How to use](https://alphapeel.readthedocs.io/en/latest/usage.html)
+
+- [Algorithm](https://alphapeel.readthedocs.io/en/latest/algorithm.html)
+
 ## Conditions of use
 
 AlphaPeel is fully and freely available for all use under the MIT License.
--- a/docs/source/algorithm.rst
+++ b/docs/source/algorithm.rst
@@ -1,3 +1,5 @@
+.. _algorithm:
+
 ==================
 Algorithm Overview
 ==================
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,13 +1,22 @@
 =========
 Changelog
 =========
-[1.3.1] - 2026-03-05
-====================
+
+.. 
+  If you are add a new entry to the changelog, please follow the format of the existing entries and 
+  add your changes to the top of the file under the [Unreleased] section. 
+  Please also include the pull request number and the GitHub username(s) of the contributor(s) who made the change.
+
+[Unreleased] - 2026-03-??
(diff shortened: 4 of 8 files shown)


XingerTang:
A preview of the new version: https://xingertang-alphapeel.readthedocs.io/en/doc-issue-230-technical-instruction/index.html

gregorgorjanc:
@XingerTang I have reviewed your changes. This looks very comprehensive - good job! I think this is ready to merge, so go ahead doing that if you don't plan any additional commits to this PR.

src/tinypeel/tinypeel.py
```diff
@@ -826,6 +830,21 @@ def getArgs():
     )
     # singleLocus_parser.add_argument('-blocksize',default=100, required=False, type=int, help='The number of markers to impute at once. This changes the memory requirements of the program.')
 
+    # special handle for version argument to allow it to be called with just -version
+    parser.add_argument(
+        "-version",
```
gregorgorjanc, line 835: Hmm, do we mention this flag in the the docs!? Please add that in the usage section, perhaps best if do that at the top where we mention the -h flag - you can add text saying that users should check what version they work with and what version the docs refer to.

XingerTang: Hi @gregorgorjanc, thanks for the review. I've pushed the commit that includes this.

XingerTang:
@gregorgorjanc I made a typo in the last commit I just pushed, that I typed ``--version`` instead of ``-version``, which latter is used in the code, also in the original issue. But at the same time, I noticed that the general convention for the multicharacter option is to use the double hyphen (see discussion in https://serverfault.com/a/387936).

There are also more discussions here (https://softwareengineering.stackexchange.com/questions/70357/command-line-options-style-posix-or-what). Suggesting using a single hyphen throughout for everything, or the double hyphen for long options.

We are currently using ``-h``, ``--help``, and all single hyphens for the remaining. 

I think maybe we can choose one of the standards and stick with it.

XingerTang:
I just made the change, fixing the typo back to ``-version``, but keep the discussion open.

gregorgorjanc:
Yeah, I thought about single and double hyphen flags/options too! Since many of our flags are quite complex/long, it will be hard to get short versions and these are tricky to remember unless users use a tool all the time (like some shell tools), so we will largely have only long versions and then using single - is fine, I guess; except -h and --help, which is ok as that gives users two entries to help. We could change --help to -help and just use single - for everything and be done with it. Happy to follow your choice;)

XingerTang:
> Yeah, I thought about single and double hyphen flags/options too! Since many of our flags are quite complex/long, it will be hard to get short versions and these are tricky to remember unless users use a tool all the time (like some shell tools), so we will largely have only long versions and then using single - is fine, I guess; except -h and --help, which is ok as that gives users two entries to help. We could change --help to -help and just use single - for everything and be done with it. Happy to follow your choice;)

I just checked the ``tinyhouse`` code, while the help message printed out suggests that we need to use ``-h`` or ``--help`` for the help message, the code actually also supports ``-help`` flag. 

The difference arises from the built-in help message for the option ``-h`` and ``--help``, which is directly generated by the ``print_help`` provided by the ``argparse`` module, and not written by us. 

I think it is possible to modify it by disabling the automatic help options support, as shown in https://docs.python.org/3/library/argparse.html#add-help 

I suggest we can keep the support for the ``--help`` as it is a general standard, but also emphasises that we can do ``-help``. 

It would require modifications in ``tinyhouse``, therefore I suggest we merge this one first, and I will open a new issue in ``tinyhouse`` to make the change.

gregorgorjanc:
I would not disable, if we can get -h and -help from our side and also --help from arg parse then even better. Given this state let's advertise in the docs that we can use -h and -help so we focus on single dash options only.

XingerTang:
> I would not disable, if we can get -h and -help from our side and also --help from arg parse then even better. Given this state let's advertise in the docs that we can use -h and -help so we focus on single dash options only.

I just added ``-help`` to the docs. 

For the built-in help options, we might need to disable them to generate a sensible help message. As the automatic generation would always show a line of 
```
options:
  -h, --help            show this help message and exit
```
At the top, regardless of whether we add ``-help``. 

We can add a new option argument with all three options written together.

gregorgorjanc:
Huh, there is always something else to do ... Is `We can add a new option argument with all three options written together.` a minor bit? If yes, go ahead and do it. If not, let's wrap up. I checked the changes and all looks well. Do you plan any more changes?

XingerTang:
> Huh, there is always something else to do ... Is `We can add a new option argument with all three options written together.` a minor bit? If yes, go ahead and do it. If not, let's wrap up. I checked the changes and all looks well. Do you plan any more changes?

It's quite easy to implement, a simple example is below. But the code should be in ``tinyhouse``, I will open the issue.

```
>>> import argparse
>>> parser = argparse.ArgumentParser(add_help=False)
>>> parser.add_argument('-h', '-help', '--help', help='display help message')
_StoreAction(option_strings=['-h', '-help', '--help'], dest='help', nargs=None, const=None, default=None, type=None, choices=None, required=False, help='display help message', metavar=None, deprecated=False)
>>> parser.print_help()
usage: [-h HELP]

options:
  -h, -help, --help HELP
                        display help message
>>> exit()
```

No other changes are planned here. I can soon squash the commits and merge this.

gregorgorjanc:
Aha, having that in `tinyhouse` is even better! Time to squash and merge this PR too @XingerTang ;)

Issue #255, opened 2026-04-28
Add a utils.py file containing the functions to calculate SER and PER for phase outputs


XingerTang:
## Goal
Add a ``utils.py`` file containing the functions to calculate SER and PER for phase outputs.
See discussion in AlphaGenes/AlphaPeel#202

## Proposed approach
Add a ``utils.py`` file under ``src/``

## Scope
``src/``

## Done when
PR open by 28 Apr

PR #256, opened 2026-04-28
Feat/issue-255-add-ser-per

XingerTang:
## Related Issue
Closes AlphaGenes/AlphaPeel#255

## What changed
- Add SER, PER calculation code to ``utils.py``

## Why this change
See AlphaGenes/AlphaPeel#255

## Notes / Risks
- Indepent file, expect no dependency issue

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -1,0 +1,115 @@
+import numpy as np
+import os
+import argparse
+
+
+def get_params():
+    param_file = os.path.join("tests", "accuracy_tests", "simulation_parameters.txt")
+    with open(param_file, "r") as file:
+        sim_params = [line.strip().split() for line in file]
+
+    params = {}
+    for param_name, param_value in sim_params:
+        params[param_name] = float(param_value)
+
+    return params
+
+
+def calc_switch_error_rate():
+    """
+    Run this function at root directory and input the path to the
+    assessed haplotype file.
+    The SER and PER_intra calculation follows the definition from:
+    https://www.cell.com/hgg-advances/fulltext/S2666-2477(25)00082-X
+    """
+
+    parser = argparse.ArgumentParser(
+        prog="calc_switch_error_rate",
+        description="Calculate switch error rate, phase error rate",
+        epilog="The SER and PER_intra calculation follows the definition from: https://www.cell.com/hgg-advances/fulltext/S2666-2477(25)00082-X",
+    )
+
+    params = get_params()
+
+    nLociAll = int(params["nLociAll"])
+    nInd = int(params["nInd"])
+
+    parser.add_argument(
+        "-true_path",
+        type=str,
+        required=False,
+        default=os.path.join(
+            "tests", "accuracy_tests", "sim_for_alphapeel_accu_test", "true-hap_0.5.txt"
+        ),
+        help="Enter the path of the true haplotype file, default is the simulation path",
+    )
+    parser.add_argument(
+        "-called_path",
+        type=str,
+        required=True,
+        help="Enter the path of the assessed haplotype file",
+    )
+
+    args = parser.parse_args()
+    true_path = args.true_path
+    called_path = args.called_path
+
+    called_file = np.loadtxt(called_path, usecols=np.arange(1, nLociAll + 1))
+    true_file = np.loadtxt(true_path, usecols=np.arange(1, nLociAll + 1))
+
+    switch_error_count = 0
+    phase_error_count = 0
+    uncalled_count = 0
+    wrong_homo_count = 0
+    true_hetero_count = 0
+    homo_count = 0
+    hetero_count = 0
+
+    for ind in range(nInd):
+        hap_p_new = called_file[ind * 2]
+        hap_m_new = called_file[ind * 2 + 1]
+        hap_p_true = true_file[ind * 2]
+        hap_m_true = true_file[ind * 2 + 1]
+
+        switched = False
+        for loci in range(nLociAll):
+            if hap_p_true[loci] == hap_m_true[loci]:
+                homo_count += 1
+            else:
+                hetero_count += 1
+            if hap_p_new[loci] == 9 or hap_m_new[loci] == 9:
+                uncalled_count += 1
+                continue
+            if hap_p_true[loci] + hap_m_true[loci] == 1:
+                if hap_p_new[loci] == hap_m_new[loci]:
+                    wrong_homo_count += 1
+                    continue
+                true_hetero_count += 1
+                if hap_p_new[loci] != hap_p_true[loci]:
+                    phase_error_count += 1
+                if (hap_p_new[loci] != hap_p_true[loci] and switched is False) or (
+                    hap_p_new[loci] == hap_p_true[loci] and switched is True
+                ):
+                    switched = not switched
+                    switch_error_count += 1
+
+    print(f"Switch error rate: {switch_error_count / (nInd * (nLociAll - 1))}")
+    print(f"Phase error (intra) rate: {phase_error_count / (nInd * nLociAll)}")
+    print(f"Uncalled rate: {uncalled_count / (nInd * nLociAll)}")
+    print(
+        f"Proportion of genotypes wrongly called as homozygote: {wrong_homo_count / (nInd * nLociAll)}"
+    )
+    print(
+        f"Proportion of genotypes correctly called as heterozygote: {true_hetero_count / (nInd * nLociAll)}"
+    )
+    print(f"Homozygote count in true genotype: {homo_count}")
+    print(f"Heterozygote count in true genotype: {hetero_count}")
+    print(f"Homo to hetero ratio: {homo_count / hetero_count}")
+
+
+def main():
+    calc_switch_error_rate()
+
+
+if __name__ == "__main__":
+    main()


docs/source/usage.rst
gregorgorjanc, line 196: Hasn't this been done already elsewhere?

gregorgorjanc, line 223: Hasn't this been done already elsewhere?

src/tinypeel/tinypeel.py
gregorgorjanc, line 532: Hasn't this been done already elsewhere?

gregorgorjanc, line 552: Hasn't this been done already elsewhere?

gregorgorjanc, line 564: Hasn't this been done already elsewhere?

gregorgorjanc, line 572: Hasn't this been done already elsewhere?

src/utils.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
+
+
+def calc_switch_error_rate():
+    """
+    Run this function at root directory and input the path to the
+    assessed haplotype file.
+    The SER and PER_intra calculation satisfies the definition from:
```
gregorgorjanc:
```suggestion
    The SER and PER_intra calculation follows the definition from:
```

src/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    https://www.cell.com/hgg-advances/fulltext/S2666-2477(25)00082-X
+    """
+    nLociAll = 2000
```
gregorgorjanc: Best to make this generic and assess from the file?

src/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    true_path = "tests/accuracy_tests/sim_for_alphapeel_accu_test/true-hap_0.5.txt"
+    file_name = input("Enter the path of the assessed haplotype file: ")
+    new_file = np.loadtxt(file_name, usecols=np.arange(1, nLociAll + 1))
+    true_file = np.loadtxt(true_path, usecols=np.arange(1, nLociAll + 1))
+
+    nInd = 1000
```
gregorgorjanc: Best to make this generic and assess from the file?

src/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+def calc_switch_error_rate():
+    """
+    Run this function at root directory and input the path to the
+    assessed haplotype file.
+    The SER and PER_intra calculation satisfies the definition from:
+    https://www.cell.com/hgg-advances/fulltext/S2666-2477(25)00082-X
+    """
+    nLociAll = 2000
+    true_path = "tests/accuracy_tests/sim_for_alphapeel_accu_test/true-hap_0.5.txt"
```
gregorgorjanc: Make it as an option so we can switch sims, but can leave as this default for ease of use. Once we start using this utility function across packages, it will likely have to move to `tinyhouse` anyway.

src/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    file_name = input("Enter the path of the assessed haplotype file: ")
+    new_file = np.loadtxt(file_name, usecols=np.arange(1, nLociAll + 1))
```
gregorgorjanc: I think it would be simpler to read the code is you use `true` (which you do) and something like `called` (instead of `new`). Apologies for nitpicking.

src/utils.py
```diff
@@ -0,0 +1,68 @@
...
+                if (hap_p_new[loci] != hap_p_true[loci] and switched is False) or (
+                    hap_p_new[loci] == hap_p_true[loci] and switched is True
+                ):
+                    switched = not switched
+                    switch_error_count += 1
+
+    print(f"Switch error rate: {switch_error_count / (nInd * (nLociAll - 1))}")
+    print(f"Phase error (intra) rate: {phase_error_count / (nInd * nLociAll)}")
+    print(f"Uncalled rate: {uncalled_count / (nInd * nLociAll)}")
+    print(f"Wrong homozygote rate: {wrong_homo_count / (nInd * nLociAll)}")
```
gregorgorjanc: I think we best divide by `homo_count`

src/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    print(f"True heterozygote rate: {true_hetero_count / (nInd * nLociAll)}")
```
gregorgorjanc: I think we best divide by `hetero_count`

src/utils.py
gregorgorjanc: Use Wrong or True for homo and hetero, but not opposiong metric, to simplify comparison

src/utils.py
gregorgorjanc: Use Wrong or True for homo and hetero, but not opposiong metric, to simplify comparison

gregorgorjanc (changes requested):
See the above comments @XingerTang

src/utils.py
XingerTang: Diving by ``homo_count`` or the ``hetero_count`` in the next comment block won't follow the definition. It would be a different metric.

Regarding the true homozygote rate or wrong heterozygote rate. I understand what we have now (wrong homozygote rate and true heterozygote rate) are quite confusing, I have displayed their relationship in the following figure. 

For the other two rates you mentioned, which seem to be composite to what we already have, but I think they are somewhat irrelevant to our focus of heterozygotes. 

Screenshot 2026-04-29 at 11 32 18

docs/source/usage.rst
XingerTang: Re-rebased the branch now, it should be fine.

src/utils.py
```diff
@@ -0,0 +1,93 @@
...
+    nInd = int(params["nInd"])
+
+    true_path = input(
+        "Enter the path of the true haplotype file (hit Enter to use default simulation path): "
+    )
+    if true_path == "":
+        true_path = os.path.join(
+            "tests", "accuracy_tests", "sim_for_alphapeel_accu_test", "true-hap_0.5.txt"
+        )
+    file_name = input("Enter the path of the assessed haplotype file: ")
```
gregorgorjanc: I assume you want to make this into an argument to the script/function so you can call it as part of a pipeline or as a command-line without user intervention - for example on compute node you can't provide input easily (maybe `./script.py < input` works, but then you already have the logic where you provide an argument.

src/utils.py
```diff
@@ -0,0 +1,93 @@
...
+    The SER and PER_intra calculation follows the definition from:
+    https://www.cell.com/hgg-advances/fulltext/S2666-2477(25)00082-X
+    """
+
+    params = get_params()
+
+    nLociAll = int(params["nLociAll"])
+    nInd = int(params["nInd"])
+
+    true_path = input(
```
gregorgorjanc: See the comment below for the same kind of input

src/utils.py
gregorgorjanc: Thanks for sharing this picture!

I appreciate the language is hard, hence lets make it clear what is calculated.

When I read `Wrong homozygote rate` my mind thinks of "the proportion of homozygotes that we called wrongly", which is `B/(A+B)`, but you compute `C/(A+B+C+D)`. I don't get why is `C/(A+B+C+D)` relevant. If it is, we should give it clearer name! `Proportion of heterozygote called as homozygote` or smth. like that `Proportion of heterozygotes called wrongly`, but this is not `C/(A+B+C+D)` - that is `Proportion of genotypes wrongly called as homozygote`.

When I read `True heterozygote rate` my mind thinks of "the proportion of heterozygotes that we called correctly", which is `D/(C+D)`, but you compute `D/(A+B+C+D)`. I don't get why is `D/(A+B+C+D)` relevant. If it is, we should give it clearer name! `Proportion of heterozygote called as heterozygote` or smth. like that `Proportion of heterozygotes called correctly`, but this is not `D/(A+B+C+D)` - that is `Proportion of genotypes correctly called as heterozygote`.

XingerTang: Your names are clearer, I have changed to them. I am more interested in the cases that have a denominator with full ``(A + B + C + D)``, because that is the denominator for the switch error rate and phase error rate, and these numbers are calculated in support of a better understanding of the values of switch error rate and phase error rate. 

The ``C/(A+B+C+D)`` is the proportion that may contribute to the switch error rate, but does not contribute because of the imputation error of genotype. This indicates that for data with better imputation accuracy, by how much the switch error rate might increase. 

The value of ``D/(A+B+C+D)`` is the upper bound of the switch error rate we computed, that the switch errors must sit inside ``D`` and is divided by ``A + B + C + D - number_of_chromsome``, which is approximately ``A + B + C + D``.

gregorgorjanc: Ah, right, yes, I see now (again!) why you want the full denominator! Indeed, best to use clearer terms. OK, resolving this chat now.

src/utils.py
```diff
@@ -0,0 +1,115 @@
...
+        "-true_path",
+        type=str,
+        required=False,
+        default=os.path.join(
+            "tests", "accuracy_tests", "sim_for_alphapeel_accu_test", "true-hap_0.5.txt"
+        ),
+        help="Enter the path of the true haplotype file, default is the simulation path",
+    )
+    parser.add_argument(
+        "-file_name",
```
gregorgorjanc: Apologies for being a pest, but shouldn't this be `-called_path` or similar (`-called_file`), to match the `-true_path` or similar (`-true_file`)?

XingerTang: You are right, consistency is important.

gregorgorjanc (approved):
Looks good now @XingerTang! Merge time;)

Issue #257, opened 2026-05-15
Remove the sequence file x_chr test for accuracy tests


XingerTang:
## Goal
The sequence file input for the sex chromosome is not implemented, and the test is using an autosomal sequence file input and the sex chromosome genotype input. The test results are misleading; therefore, remove the tests for now.

## Proposed approach
Modify ``tests/accuracy_tests/run_accu_test.py``

## Scope
``tests/accuracy_tests/run_accu_test.py``

## Done when
Soon.

PR #261, opened 2026-05-18
remove sequence file#257

AprilYUZhang:
see issue #257


XingerTang (approved):

XingerTang:
This is a very simple change that only removes the two lines in the test files.

Issue #260, opened 2026-05-15
Error with `-est_start_alt_allele_prob`


RosCraddock:
## Goal
Fix bug so that `-est_start_alt_allele_prob` functions correctly to match previous accuracy tests performed in issue #41. This bug was introduced when updating the code to allow multiple metafounders within a single individual. The anterior probabilities were not updated as the if statement was not triggered. 
Additionally, will check the function of `est_alt_allele_prob` too.

## Proposed approach
Correct if statement so triggered in peelingUpdates.updateMafs() and modify so anterior probs are updated

## Scope
What files/modules are affected?
peelingUpdates.py

## Done when
Bug is resolved.

PR #263, opened 2026-05-18
Fix issue #260

RosCraddock:
### Related Issue
Issue #260 

### What changed
peelingUpdates.py - added code to allocate anterior probabilities from estimated allele frequencies using up to two metafounders per individual. The if statement previously was not triggered, but is working now.
run_func_test.py - adapted test_est to capture this error in future (previous functional test passed regardless of whether est_start_alt_allele_prob was used or not).
run_accu_test.py - added two tests for 'est_alt_allele_prob' without metafounders

### Why this change?
To fix the error described in issue #260 and catch it if to occur again in future

### Notes
Minor changes


tests/functional_tests/run_func_test.py
```diff
@@ -394,6 +393,54 @@ def test_est(self):
...
+        # test est_start_alt_allele_prob functionality
+        self.test_cases = "est_start_alt_allele_prob"
+        self.input_files = ["geno_file", "ped_file"]
+        self.input_file_depend_on_test_cases = self.input_files
+        self.arguments["est_start_alt_allele_prob"] = None
+        self.arguments["alt_allele_prob"] = None
+        self.output_file_prefix = f"est.{self.test_cases}"
+
+        self.generate_command()
+        print(self.command)
```
XingerTang: Remove this line, and it will be all good. 😊

XingerTang (commented):
Looks very good to me

tests/functional_tests/run_func_test.py
RosCraddock: By this line, do you mean 403 or 405?

XingerTang: Line 405, maybe also remove lines 377 and 1237.

XingerTang:
Some new plots after this change:

Nice to see nothing is fully overlapped


-----
Edited: the graphs are wrong, therefore deleted.

tests/functional_tests/run_func_test.py
RosCraddock: Done!

gregorgorjanc:
@XingerTang I will leave you to merge this one (after any additionally needed edits are sorted). Would be good to see the new imputation quality plots.

XingerTang:
Everything here looks good to me. 

For the ImputationQuality plot, the ImputationQuality software has a misleading option name; I was using the wrong option for all the analyses by yesterday. The MAF was not around 0.5 for all of the cases assessed, and the values of the evaluated metrics are very different now. 


After the recalculation, the MAF for both ``est_start_alt_allele_prob`` and ``est_alt_allele_prob``options is plotted below. We can see that the estimations are quite accurate:
newplot (5)

Just for reference for comparison, the following are the MAF plots for ``multi`` and ``single`` with no estimation. There is clearly a larger gap between the true and the outputs:
newplot (6)

While the behaviour of the metrics is quite bizarre, for ``r2_BEAGLE``, there are a few very low values for some of the loci after the estimation: 

newplot (8)

If we zoom in and restrict the range to the sensible part, we can see that the estimation does not guarantee to improve the metric.
newplot

The ``r2_MACH``shows a similar trend as ``r2_BEAGLE``. While for the ``Iam_chance``, estimation would improve the values of the metric for ``multi`` but decrease for ``single``. 
newplot (1)

For metafounders, some of the metrics are:
newplot (2)
newplot (3)
newplot (4)

gregorgorjanc:
> If we zoom in and restrict the range to the sensible part, we can see that the estimation does not guarantee to improve the metric.

Yes, that's why by default these parameters are kept fixed and not estimated from the data and we provide this as an option.

> While the behaviour of the metrics is quite bizarre, for r2_BEAGLE, there are a few very low values for some of the loci after the estimation:

It would be useful to understand what happens with these outliers - clearly the estimation and expectations mismatch - I assume this is a fault of the metric rather than the estimation or just plain inference, but would be good to understand where it's coming from. This is obviously independent of the contents of this PR.

XingerTang:
From the [Supplementary material](https://gitlab.gwdg.de/kolja.thormann1/imputationquality/-/blob/main/Supplement_V4.3__final_.pdf?ref_type=heads), the definition of Beagle $R^2$ is given by

Screenshot 2026-05-20 at 16 53 43

The software sets the value to be -9 when the value of the denominator is not positive.

Issue #252, opened 2026-04-22
Modify output logic of ``geno_error_prob``, ``rec_prob``, ``seq_error_prob``


XingerTang:
## Goal
Cleanup default outputs. 
See https://github.com/AlphaGenes/AlphaPeel/issues/249

Probably would handle the remove ``no_param`` part of https://github.com/AlphaGenes/AlphaPeel/issues/247 as well.

## Proposed approach
- [ ] Remove the following in ``tinypeel.py``
     ```    
        if not args.no_param:
          PeelingIO.writeOutParamaters(peelingInfo)
  ```
- [ ] Add new options and relevant output functions that control the parameter outputs.
- [ ] Add relevant tests to check if the options are working
- [ ] Update docs correspondingly

## Scope
``src/tinypeel/tinypeel.py``, ``src/tinypeel/Peeling/PeelingIO.py``, ``tests/``, ``docs/``

## Done when
PR open by Apr 24

gregorgorjanc:
Yes, remove the `-no_param` as part of this change as we flipped the logic (output params when they are estimated).

PR #268, opened 2026-05-22
Refactor/issue 252 param output

XingerTang:
## Related Issue
Closes AlphaGenes/AlphaPeel#252

## What changed
- Remove option ``-no_param``
- Add options ``-estimated_geno_error_prob``, ``-estimated_seq_error_prob``, ``-rec_prob``
- Relevant updates in docs and tests

## Why this change
- Output logic consistency

## Notes / Risks
- Minor changes


src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -93,11 +93,18 @@ def writeOutParamaters(peelingInfo):
...
+            args.out_file + ".geno_error_prob.txt", peelingInfo.genoError, fmt="%f"
+        )
+    if args.estimated_seq_error_prob:
+        np.savetxt(
+            args.out_file + ".seq_error_prob.txt", peelingInfo.seqError, fmt="%f"
+        )
+    if args.rec_prob:
+        np.savetxt(
+            args.out_file + ".rec_prob.txt", np.empty((1, 1)), fmt="%f"
+        )  # not be realized, just as a placeholder
```
gregorgorjanc: I assume this comment should be removed?

gregorgorjanc: Or does this mean that we can not write out the used recombination rate?

src/tinypeel/tinypeel.py
```diff
@@ -676,19 +676,31 @@ def getArgs():
         "-seg_prob",
         action="store_true",
         required=False,
-        help="Flag to enable writing out the segregation probabilities.",
+        help="Ourpur segregation probabilities (see format details in the docs).",
```
gregorgorjanc:
```suggestion
        help="Output segregation probabilities (see format details in the docs).",
```

src/tinypeel/tinypeel.py
```diff
@@ -676,19 +676,31 @@ def getArgs():
...
         action="store_true",
         required=False,
-        help="Flag to suppress writing the model parameter files.",
+        help="Output estimated genotype error probabilities (see format details in the docs).",
+    )
+    output_parser.add_argument(
+        "-estimated_seq_error_prob",
+        action="store_true",
+        required=False,
+        help="Flag to enable writing out the estimated sequence error probabilities.",
```
gregorgorjanc: Use the same style as with others

src/tinypeel/tinypeel.py
```diff
@@ -676,19 +676,31 @@ def getArgs():
...
+    )
+    output_parser.add_argument(
+        "-rec_prob",
+        action="store_true",
+        required=False,
+        help="Flag to enable writing out the recombination probabilities (NOT IMPLEMENTED YET).",
```
gregorgorjanc: Use the same style as with others.

I think we can write out what is internally used, but appreciate that it is not estimated from the data.

So, should we then remove "estimated" from these outputs to be consistent with this ... let me get back to this after reviewing all the flag names ...

docs/source/usage.rst
```diff
@@ -175,6 +175,11 @@ Output options
                             Value(s) less than 1/2 are replaced by 1/2.
       -seg_prob             Output :ref:`segregation probabilities <seg_prob_file_format>`.
       -pheno_prob           Output :ref:`phenotype probabilities <pheno_prob_file_format>`.
+      -estimated_geno_error_prob
```
gregorgorjanc: @XingerTang We already have these in the Peeling parameters section - see https://github.com/AlphaGenes/AlphaPeel/blob/674ae6b79a542c47a1cad8f08dbb1d3fb9c66937/docs/source/usage.rst?plain=1#L273, so follow that template.

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -93,11 +93,18 @@ def writeOutParamaters(peelingInfo):
     """
     args = InputOutput.args
 
-    np.savetxt(args.out_file + ".geno_error_prob.txt", peelingInfo.genoError, fmt="%f")
-    np.savetxt(args.out_file + ".seq_error_prob.txt", peelingInfo.seqError, fmt="%f")
-    np.savetxt(
-        args.out_file + ".rec_prob.txt", np.empty((1, 1)), fmt="%f"
-    )  # not be realized, just as a placeholder
+    if args.estimated_geno_error_prob:
```
gregorgorjanc: Sync with https://github.com/AlphaGenes/AlphaPeel/blob/674ae6b79a542c47a1cad8f08dbb1d3fb9c66937/docs/source/usage.rst?plain=1#L273

gregorgorjanc: And the same below

src/tinypeel/Peeling/PeelingIO.py
gregorgorjanc: Yes, it's not implemented atm.

gregorgorjanc:
```suggestion
        )  # not implemented atm, just a placeholder
```

src/tinypeel/tinypeel.py
```diff
@@ -676,19 +676,31 @@ def getArgs():
...
     output_parser.add_argument(
         "-pheno_prob",
         action="store_true",
         required=False,
-        help="Flag to enable writing out the phenotype probabilities.",
+        help="Output phenotype probabilities (see format details in the docs).",
     )
     output_parser.add_argument(
-        "-no_param",
+        "-estimated_geno_error_prob",
```
gregorgorjanc: Sync with https://github.com/AlphaGenes/AlphaPeel/blob/674ae6b79a542c47a1cad8f08dbb1d3fb9c66937/docs/source/usage.rst?plain=1#L273

gregorgorjanc: And below

src/tinypeel/tinypeel.py
gregorgorjanc: OK, we will follow the same style as with others - if we don't estimate it, then we don't write it out.

tests/functional_tests/run_func_test.py
```diff
@@ -441,8 +442,14 @@ def test_est(self):
 
     def test_no(self):
         """
-        Check to make sure the no_dosage, seg_prob, no_param, phased_geno_prob
-        flags work.
+        Check to make sure the following flags work:
+        - no_dosage,
+        - seg_prob,
+        - estimated_geno_error_prob,
```
gregorgorjanc: Sync with https://github.com/AlphaGenes/AlphaPeel/blob/674ae6b79a542c47a1cad8f08dbb1d3fb9c66937/docs/source/usage.rst?plain=1#L273

gregorgorjanc (changes requested):
@XingerTang I had a look at this PR - make sure you sync flag names with https://github.com/AlphaGenes/AlphaPeel/blob/674ae6b79a542c47a1cad8f08dbb1d3fb9c66937/docs/source/usage.rst?plain=1#L273 as we have already picked the names of these flags/args;)

docs/source/usage.rst
XingerTang: The options in the estimation of model parameters section would enable the estimation of the parameters, but not output them. 
I am wondering if we should separate the two options or combine them together.

XingerTang: I am using ``estimated`` because ``geno_error_prob`` is for the input.

gregorgorjanc: I suggest that if we estimate params, then we likely want to know what the estimates were, so let's output by default. That's how we have written the docs;) I see the point on having the option to output or not, but let's not introduce another set of command line args just for that, I reckon.

docs/source/usage.rst
```diff
@@ -175,11 +175,9 @@ Output options
                             Value(s) less than 1/2 are replaced by 1/2.
       -seg_prob             Output :ref:`segregation probabilities <seg_prob_file_format>`.
       -pheno_prob           Output :ref:`phenotype probabilities <pheno_prob_file_format>`.
-      -estimated_geno_error_prob 
-                            Output :ref:`genotype error probabilities <geno_error_prob_file_format>`.
-      -estimated_seq_error_prob
-                            Output :ref:`sequence error probabilities <seq_error_prob_file_format>`.
-      -alt_allele_prob      Output :ref:`alternative allele probabilities <alt_allele_prob_file_format>`.
+      -alt_allele_prob      Output :ref:`alternative allele probabilities for each metafounder <alt_allele_prob_file_format>`.
```
gregorgorjanc: Best to remove `for each metafounder` as we deal with that detail in the file format. The idea is that we deal with one piece of information in only one place, ideally. Otherwise it becomes hard to track and keep in sync what is documented where.

docs/source/usage.rst
```diff
@@ -282,9 +280,11 @@ Peeling parameters
       -est_alt_allele_prob  Estimate :ref:`alternative allele probabilities <alt_allele_prob_file_format>`
                             after each peeling cycle.
       -est_geno_error_prob  Estimate :ref:`genotype error probability <geno_error_prob_file_format>`
-                            after each peeling cycle.
+                            after each peeling cycle. 
+                            Output :ref:`genotype error probabilities <geno_error_prob_file_format>`.
```
gregorgorjanc: I don't think we need `Output :ref:...` since we already link to that at the beginning. But I think you are raising a good point here about the "output" so I suggest this wording: `Estimate and output ...`. But do this for all estimation options so we sync them all (the current proposal deviates from uniform style).

docs/source/usage.rst
```diff
@@ -282,9 +280,11 @@ Peeling parameters
...
       -est_seq_error_prob   Estimate :ref:`sequence error probability <seq_error_prob_file_format>`
-                            after each peeling cycle.
+                            after each peeling cycle. 
+                            Output :ref:`sequence error probabilities <seq_error_prob_file_format>`.
```
gregorgorjanc: As above

src/tinypeel/tinypeel.py
```diff
@@ -787,13 +775,13 @@ def getArgs():
         "-est_geno_error_prob",
         action="store_true",
         required=False,
-        help="Estimate genotype error probability after each peeling cycle.",
+        help="Estimate genotype error probability after each peeling cycle. Output estimated genotype error probabilities (see format details in the docs).",
```
gregorgorjanc: `Estimate and output ...`

src/tinypeel/tinypeel.py
```diff
@@ -787,13 +775,13 @@ def getArgs():
...
     )
     peeling_control_parser.add_argument(
         "-est_seq_error_prob",
         action="store_true",
         required=False,
-        help="Estimate sequence error probability after each peeling cycle.",
+        help="Estimate sequence error probability after each peeling cycle. Output estimated sequence error probabilities (see format details in the docs).",
```
gregorgorjanc: `Estimate and output ...`

gregorgorjanc (changes requested):
One more iteration @XingerTang and this will be done.

docs/source/usage.rst
XingerTang: I would update the help message. For the ``est_alt_allele_prob`` and ``est_pheno_penetrance_prob``, their outputs are controled separately by ``alt_allele_prob`` and ``pheno_penetrance_prob`` options. Therefore, using the estimation does not necessarily output the corresponding files.

gregorgorjanc: Good inconsistency catch @XingerTang, that is, we are not consistent with `est_alt_allele_prob` and `est_pheno_penetrance_prob` in a sense the these only estimate, but don't output the estimates (while other est options do both). I suggest that we modify this behaviour and estimate & output like with other `est_*` options and remove the `alt_allele_prob` and `pheno_penetrance_prob` output options for the sake of consistency (and less options)! OK?

XingerTang: There are test cases that we want to ensure that the alternative allele probabilities input are properly used without the estimation. Should we ignore these test cases or we also allow alt_allele_prob output for the cases we have alt_allele_prob input? (same apply to other parameter file inputs)

gregorgorjanc: Darn. Ok, what about this proposal. When we use an `est_*` flag we estimate and output by default. This makes sense and can be implemented for all est flags. Since we can read in the alt allele probs, it’s good that we allow for the `alt_allele_prob` output flag, as you say. I assume the same holds for the `pheno_penetrance_prob` output?

XingerTang:
Now, all the estimation options would output the corresponding estimated parameter file. The options ``alt_allele_prob`` and ``pheno_penetrance_prob`` can output the corresponding parameter files regardless of whether the estimation is implemented, while the genotype error probabilities and sequence error probabilities do not have these independent output options.

src/tinypeel/tinypeel.py
```diff
@@ -763,31 +763,31 @@ def getArgs():
...
         required=False,
-        help="Estimate alternative allele probabilities after each peeling cycle.",
+        help="Estimate after each peeling cycle and output alternative allele probabilities (see format details in the docs).",
     )
     peeling_control_parser.add_argument(
         "-est_geno_error_prob",
         action="store_true",
         required=False,
-        help="Estimate genotype error probability after each peeling cycle.",
+        help="Estimate after each peeling cycle and output genotype error probability (see format details in the docs).",
```
gregorgorjanc: Do we estimate one or more geno error rate(s)? If one then the text is right, if more (one pre locus), then use `probabilities` (similarly elsewhere0

gregorgorjanc: For `seq_err_...` we use plural, for example. Check and sync.

tests/functional_tests/run_func_test.py
```diff
@@ -441,42 +441,60 @@ def test_est(self):
 
     def test_no(self):
         """
-        Check to make sure the no_dosage, seg_prob, no_param, phased_geno_prob
-        flags work.
+        Check to make sure the following flags output the correct files:
```
gregorgorjanc, line 444: Do we need to add alt_allele_prob here too or not?

gregorgorjanc (changes requested):
Nic convergence @XingerTang ;) Minor bits left and we are close to merging this in!

gregorgorjanc (approved):
I thinks this looks good to merge in @XingerTang!

Issue #271, opened 2026-05-27
Add option to control output rounding digits


XingerTang:
## Goal
Add an option to control output rounding digits
Relate to issue AlphaGenes/AlphaPeel#68

## Proposed approach
Add an option for rounding digits for all floating-point outputs.
Update relevant tests and docs.

## Scope
```
docs/source/usage.rst
src/tinypeel/Peeling/PeelingIO.py
src/tinypeel/tinypeel.py
tests/functional_tests/run_func_test.py
ests/functional_tests/test_est/true-alt_allele_prob-est_start_alt_allele_prob.txt
```

## Done when
PR open immediately.

PR #272, opened 2026-05-27
Feat/issue 271 rounding digits

XingerTang:
## Related Issue
Closes AlphaGenes/AlphaPeel#271

## What changed
See AlphaGenes/AlphaPeel#271

## Why this change
Allow more precise outputs.

## Notes / Risks
- f-strings (formatted string literals) were introduced in Python 3.6, adding the requirement to the ``pyproject.toml``.
- f-strings are used in this pull request, but it is already used quite widely in AlphaPeel before this pull request


gregorgorjanc:
@XingerTang As discussed in one of the posts/PR/... would it not be best if we:
1) only limit digits only for "quadratic" outputs (dosages, genotype probs etc.) as that's where storage bulks up and
2) output full precision for the "linear" outputs (allele probs, recombination probs, etc.)

Otherwise the approach you are making is clear and good!

XingerTang:
After I remove all the formatting of the alt_allele_prob, I get: 
```
MF_1
7.142737507820129395e-01
2.857262790203094482e-01
7.142737507820129395e-01
```
I find it quite difficult to read. I prefer still having the rounding on.

gregorgorjanc:
I agree it's lots of numbers to look at! But do we really look at them that much? I guess we do at least in testing, right? I don't have strong thoughts here. I just thought that we would want more precision in these hyper parameter outputs; the bit that I have particularly in mind is rounding to zero, which is why we started to discuss imposing MAF limits, and I assume the same will appear with other parameters. I am aware we are discussing output here only, however this output can then be read back in as starting values in estimation or as fixed values, which is where zeros can creep in etc. Maybe I am splitting hairs here again!? This read-in strategy could be a nice way to save on estimation time yet obtain more accurate results!? Worth mentioning in docs somewhere?


@RosCraddock what do you think from a users perspective?

RosCraddock:
I agree with @XingerTang, that it is harder to digest. However, as you say @gregorgorjanc, I can see the high precision to be beneficial if the output is to be read back in and for ensuring reproducibility. Even when rounded, these parameters can still be hard to read, so they are often plotted. So, I think the benefit of full precision outweighs the readability trade-off.

docs/source/usage.rst
```diff
@@ -196,6 +196,7 @@ Output options
       -out_id_only          Suppress output for individuals not present in
                             the file specified with -out_id_order. It also suppresses
                             "dummy" individuals.
+      -out_digits           Specify the number of digits to round to in the outputs except ``alt_allele_prob``, ``geno_error_prob`` and ``seq_error_prob``. Default: 4.
```
gregorgorjanc:
```suggestion
      -out_digits           Specify the number of digits to round the outputs.
                            Does not apply to outputs from ``alt_allele_prob``, 
                            ``geno_error_prob``, and ``seq_error_prob``.
                            Default: 4.
```

src/tinypeel/Peeling/PeelingIO.py
```diff
@@ -149,7 +150,9 @@ def writePhenoPenetrance(pedigree):
     """
     args = InputOutput.args
     np.savetxt(
-        args.out_file + ".pheno_penetrance.txt", pedigree.phenoPenetrance, fmt="%.2f"
+        args.out_file + ".pheno_penetrance.txt",
```
gregorgorjanc, line 153: Hmm, this is population based penetrance, right? If so, then rounding should be skipped to and option named in `out_digits`.

src/tinypeel/tinypeel.py
```diff
@@ -545,6 +545,13 @@ def get_output_options():
         required=False,
         help='Suppress output for individuals not present in the file specified with -out_id_order. It also suppresses "dummy" individuals.',
     )
+    parse_dictionary["out_digits"] = lambda parser: parser.add_argument(
+        "-out_digits",
+        default=4,
+        type=int,
+        required=False,
+        help="Specify the number of digits to round to in the outputs except ``alt_allele_prob``, ``geno_error_prob`` and ``seq_error_prob``. Default: 4.",
```
gregorgorjanc: Sync with above text suggestion

gregorgorjanc (changes requested):
Minor tweaks left

gregorgorjanc (approved):
Looks great @XingerTang! Merge away;)

Issue #231, opened 2026-03-17
Document X chromosome outputs


gregorgorjanc:
## Goal
X chromosome inputs are described in https://github.com/AlphaGenes/AlphaPeel/commit/36af2da38d0cdaae039a428f4c394a8db8151e9e (males 0,1 and females 0,1,2 genotypes and sex in pedigree (0 for males and 1 for females)), but we also need to document what the output mean - in particularly for the males.

## Proposed approach
Document for each output file how to interpret the values.

## Scope
`docs/source/usage.rst` - now that I have written this, I wonder if we should split that file into `docs/source/usage.rst` and `docs/source/outputs.rst`!?

## Done when
When we have clear description of each output file for X chromosome.

## Will be able to do it once PR https://github.com/AlphaGenes/AlphaPeel/pull/226 is merged in

Issue #167, opened 2024-06-06
If input data has phased 


AprilYUZhang:

gregorgorjanc:
Sometimes we get haplotypes from some other tool. It would be nice to be able to read in these phased genotypes and leverage that information. This is a low priority.

XingerTang:
@gregorgorjanc @AprilYUZhang 

We don't need to build a new functionality, `AlphaPeel` supports the `penetrance` input which takes phased genotype probabilities input. The `penetrance` input option works just fine, yet it is not in the documentation (there is an old issue about this #74).

`AlphaPeel` also has a phase data input option `hap_file` (originally `phasefile`) whilst it is also not in the documentation, however, the input phased information is only used to fill in the missing genotype information before the calculation of the genotype probabilities, and the phased information is not being used in the actual peeling process.

AprilYUZhang:
Nice, and then I will test if it still works and compare the input format and output format from other software.

Issue #74, opened 2023-09-07
Add Penetrance File in Documentation


XingerTang:
The penetrance file input option is working well, but it neither has a help message (it is commented in the source code) nor any explanation in the documentation.
We need to review the use of the penetrance file. If we don't need it anymore, then we can remove it completed from the source code. Or, if we find it useful, then we need to write a help message and corresponding information in the documentation.

Issue #73, opened 2023-09-07
Phase File Input and Output


XingerTang:
For input:

We have an option for the phasefile input and a lot of tests written previously are still using the phasefile input. However, for AlphaPeel, the use of the external phasefile is not supported and the phase information will be translated into genotype probabilities. So:

- If we are going to use the phasefile as an input option, we may need to move/copy the section in the documentation explaining the format of phasefile from the output file format section to the input file format section, and add the help message for the input option `-phasefile` to the documentation as well
- If we want to discard the option, we might need to remove that input option for AlphaPeel. 

For output:

The output option `-call_phase` for the phase file is supported by AlphaPeel as long as the calling threshold is provided. However, the description of the option is not included in the documentation. So we might need to copy the help message of `-call_phase` to the documentation and write a description about how to use it.

gregorgorjanc:
@XingerTang as part of this issue, I would like to clarify/unify the use of terms haplotype, phase, phased genotype etc. in AlphaPeel as it can get confusing for users. Can you please collect what terms are used in what context so that we can create a plan what to call what and stick to that plan, particularly in the documentation and hence AlphaPeel options.

PR #278, opened 2026-06-03
Refactor/clarify-input-files

XingerTang:
## Related Issue
Closes https://github.com/AlphaGenes/AlphaPeel/issues/73
Closes https://github.com/AlphaGenes/AlphaPeel/issues/74
Closes https://github.com/AlphaGenes/AlphaPeel/issues/127
Closes https://github.com/AlphaGenes/AlphaPeel/issues/167
Closes https://github.com/AlphaGenes/AlphaPeel/issues/231

## What changed
- Update code for haplotype and penetrance inputs
- Remove the reference option, it is not supported by AlphaPeel
- Update documentation to include information of haplotype and penetrance inputs
- Other minor changes

## Why this change
- To clarify the usage of the input options.

## Submodule changes
- Submodule repository: AlphaGenes/tinyhouse
- PR: https://github.com/AlphaGenes/tinyhouse/pull/38
- Commit: 8d31a5268b0d5b11cc1837f80132ff624c15ef40
- Why needed: The hap file message is misleading now, given the option name is now ``hap_file`` instead of ``phasefile``. 

## Notes / Risks
- No major code change


docs/source/usage.rst
```diff
@@ -76,6 +79,9 @@ Input options
                           Sequence error :ref:`probability <prob_freq_rate>`.
                           Must not be 0.
                           Default: 0.001.
+      -penetrance [PENETRANCE ...]
```
gregorgorjanc: Aha, this is the "hidden" non-documented option - thanks for brining it up into docs @XingerTang! We should be clearer about the name in line with the other terms we have used for the other input files - would you do this as part to this PR or follow up? What arg name would you pick?

docs/source/usage.rst
```diff
@@ -39,6 +39,9 @@ Input options
       -pheno_file [PHENO_FILE ...]
                           Phenotype file(s)
                           (:ref:`see format details <pheno_file_format>`).
+      -hap_file [HAP_FILE ...]
```
gregorgorjanc: Since this is a genetic information file I think it best if we put it before the -pheno_file, but best to add it after the -seq_file since -hap_file is a derived file, not primary data.

docs/source/usage.rst
```diff
@@ -76,6 +79,9 @@ Input options
                           Sequence error :ref:`probability <prob_freq_rate>`.
                           Must not be 0.
                           Default: 0.001.
+      -penetrance [PENETRANCE ...]
+                          Optional external :ref:`penetrance <penetrance_file_format>` file(s). 
+                          This will overwrite the default penetrance values.
```
gregorgorjanc: Do we need to say more as we have individual penetrance probs and model parameters so it gets confusing ...

docs/source/usage.rst
```diff
@@ -87,9 +93,16 @@ one or more genomic data files to run the analysis.
 
 |Software| supports the following genomic data files:
 genotype file in the AlphaGenes format (``-geno_file``),
-sequence allele read count file in the AlphaGenes format (``-seq_file``), and
+sequence allele read count file in the AlphaGenes format (``-seq_file``),
+haplotype file in the AlphaGenes format (``-hap_file``), and
 binary Plink file (``-plink_file``).
 
+|Software| supports genotype probabilities input in the AlphaGenes format (``-penetrance``). 
+The ``-penetrance`` input assumes the genotype probability data is error-free.
```
gregorgorjanc: Are these individual geno probs stored in penetrance probs? I think so, but check. If yes, then we should probably write something like `The ``-penetrance`` input is assumed to contain error-free phased? genotype probabilities. Hence, you should not ...`

docs/source/usage.rst
```diff
@@ -166,7 +179,7 @@ Output options
                             Value(s) less than 1/3 are replaced by 1/3.
       -geno_prob            Output :ref:`genotype probabilities <geno_prob_file_format>`.
       -phased_geno_prob     Output :ref:`phased genotype probabilities <phased_geno_prob_file_format>`.
-      -hap                  Call and output :ref:`haplotypes <hap_file_format>`.
+      -hap                  Call and output :ref:`haplotypes <haplotype_file_format>`.
```
gregorgorjanc, line 189: I think you want to revert this back since the tag has the short form and I think we use short form to the tags.

gregorgorjanc: Aha, there are two, ugh, tricky. Let me check this out.

docs/source/usage.rst
```diff
@@ -549,6 +562,65 @@ Example with four individuals and their phenotypes for a binary trait:
...
+  id1 0 1 1 0 # Paternal haplotype
+  id1 0 1 1 0 # Maternal haplotype
+  id2 1 0 1 0
+  id2 0 1 0 1
+  id3 0 1 1 0
+  id3 1 0 1 0
+  id4 0 1 1 0
+  id4 0 1 0 0
+
+The X chromosome functionality for haplotype input is not tested.
```
gregorgorjanc:
```suggestion
The X chromosome functionality for haplotype input is not tested, so do not use it. 
```

gregorgorjanc: Best to open issue about this so we track it.

docs/source/usage.rst
```diff
@@ -549,6 +562,65 @@ Example with four individuals and their phenotypes for a binary trait:
...
+
+The X chromosome functionality for haplotype input is not tested. 
+
+.. _penetrance_file_format:
+
+Penetrance file
+===============
+
+This file has four lines with *phased genotype probabilities* for each individual.
+The four lines correspond to probabilities for
```
gregorgorjanc: The example below has 3 lines per ind;)

docs/source/usage.rst
```diff
@@ -549,6 +562,65 @@ Example with four individuals and their phenotypes for a binary trait:
   id3 1
   id4 0
 
+.. _hap_file_format:
```
gregorgorjanc, line 547: Move above pheno file

docs/source/usage.rst
```diff
@@ -549,6 +562,65 @@ Example with four individuals and their phenotypes for a binary trait:
...
+
+Haplotype file
+==============
+
+This file has two lines with
+*observed phased haplotypes* for each phased genotyped individual.
```
gregorgorjanc:
```suggestion
This file has two lines with
*observed phased haplotypes* for each phased genotyped individual;
possibly as an external reference haplotype library or
from previous analyses.
```

@XingerTang what are your thoughts on this suggestion?

docs/source/usage.rst
```diff
@@ -549,6 +562,65 @@ Example with four individuals and their phenotypes for a binary trait:
...
+  id4 0 1 0 0
+
+The X chromosome functionality for haplotype input is not tested. 
+
+.. _penetrance_file_format:
+
+Penetrance file
+===============
+
+This file has four lines with *phased genotype probabilities* for each individual.
```
gregorgorjanc:
```suggestion
This file has four lines with
*phased genotype probabilities* for each individual;
possibly as an external probabilistic reference haplotype library or
from previous analyses.
```

@XingerTang would this work / be correct?

gregorgorjanc (changes requested):
@XingerTang see my comments

docs/source/usage.rst
XingerTang: I was also thinking about it, ``penetrance_file`` or ``geno_prob_file`` could be a good choice.

docs/source/usage.rst
XingerTang: I put it here just because the options groups in the code are decided so. But I think it should be fine if I move this option to the group of other input file options.

docs/source/usage.rst
XingerTang: ``-penetrance`` input assumes ``0`` error and goes directly to the genotype probabilities, while overwriting other inputs if there is any overlap in the data. I think it would be quite risky to use the reference haplotype library here.

docs/source/usage.rst
XingerTang: This is safer to do than ``penetrance`` as it will consider the other inputs alongside. But I don't think it is designed to be used as the reference haplotype library.

docs/source/usage.rst
XingerTang: Issue is opened: https://github.com/AlphaGenes/AlphaPeel/issues/279

docs/source/usage.rst
gregorgorjanc: I think geno_prob_file would indeed be the best. Remind me please, are these “penetrances” phased or non-phased? I think they are phased, but could be wrong. I am asking as we have two geno prob outputs and it would be good to sync this arg name with that: geno_prob and phased_geno_prob. So should this arg be geno_prob_file or phased_geno_prob_file?

docs/source/usage.rst
gregorgorjanc: If this is input for individuals then best to move it up with other inputs for individuals. We should be very clear about the overwriting in a clear way (here or elsewhere).

docs/source/usage.rst
```diff
@@ -87,9 +93,16 @@ one or more genomic data files to run the analysis.
 
 |Software| supports the following genomic data files:
 genotype file in the AlphaGenes format (``-geno_file``),
-sequence allele read count file in the AlphaGenes format (``-seq_file``), and
+sequence allele read count file in the AlphaGenes format (``-seq_file``),
+haplotype file in the AlphaGenes format (``-hap_file``), and
 binary Plink file (``-plink_file``).
 
+|Software| supports genotype probabilities input in the AlphaGenes format (``-penetrance``).
```
gregorgorjanc: Sync arg name and there is “phased?” that needs polish

gregorgorjanc:
@XingerTang this PR is a bit confusing to me because we have two sections for a haplotype file, but one is used for input (where only some individuals can be present) and one for output (where all individuals will be present). So the file format is the same, but there is the nuance of possibly few individuals on input and all individuals on output. Then we have also the input for penetrances.

So, for the haplotypes, I think it best if we have one haplotype file format section (to avoid repetition and drift in essentially the same text) and we describe that when used on input this file doesn’t need to have all individuals present. I assume we also need to be clear what happens in the code with these inputted haplotypes - are they converted to “penetrances” or what?

I gave a comment on penetrances but then got stuck in the haplotypes, so lets address haplotypes first and then we sort out the penetrances.

XingerTang:
> @XingerTang this PR is a bit confusing to me because we have two sections for a haplotype file, but one is used for input (where only some individuals can be present) and one for output (where all individuals will be present). So the file format is the same, but there is the nuance of possibly few individuals on input and all individuals on output. Then we have also the input for penetrances.
> 
> So, for the haplotypes, I think it best if we have one haplotype file format section (to avoid repetition and drift in essentially the same text) and we describe that when used on input this file doesn’t need to have all individuals present. I assume we also need to be clear what happens in the code with these inputted haplotypes - are they converted to “penetrances” or what?
> 
> I gave a comment on penetrances but then got stuck in the haplotypes, so lets address haplotypes first and then we sort out the penetrances.

We also have genotype input and output formats in two sections. I think it is acceptable to have some level of repetition. So when the user navigates the input format sections and the output format sections separately, they can always find a complete list of formats. 

Also, the difference between inputs and outputs for the haplotype file includes the X chromosome as well as the calling threshold stuff.

docs/source/usage.rst
XingerTang: These are phased, yes, it would be better if we use ``phased_geno_prob_file``.

gregorgorjanc: Looking at all the options, including the outputs, where we have `-phased_geno_prob` to write out phased geno probs into a file, then indeed we should call this input file option as `phased_geno_prob_file`.

docs/source/usage.rst
gregorgorjanc: OK, I see you followed the previous format with geno/genotype_file_format, which is good.

docs/source/usage.rst
gregorgorjanc: Thinking a bit more, this should likely be:

```
This file has two lines with
*observed haplotypes* for each genotyped and phased individual;
possibly from previous analyses.
```

@XingerTang let's talk next week how this input is actually used - can you point me in the code where it is used?

docs/source/usage.rst
gregorgorjanc: Taken together this should likely be:

```
Phased genotype probabilities file
==================================

This file has four lines with
*phased genotype probabilities* for each genotyped and phased individual;
possibly from previous analyses.
The four lines respectively correspond to probabilities for
``aa``, ``aA``, ``Aa``, and ``AA`` phased genotypes.
The file does not need to include all individuals present in other files.
The first value in each line is the individual ID.
The remaining values are phased genotype probabilities at each locus.
These values are assumed without an error and
used directly as individual's phased genotype probability state.
```

docs/source/usage.rst
```diff
@@ -87,9 +93,16 @@ one or more genomic data files to run the analysis.
...
 |Software| supports the following genomic data files:
 genotype file in the AlphaGenes format (``-geno_file``),
-sequence allele read count file in the AlphaGenes format (``-seq_file``), and
+sequence allele read count file in the AlphaGenes format (``-seq_file``),
+haplotype file in the AlphaGenes format (``-hap_file``), and
 binary Plink file (``-plink_file``).
 
+|Software| supports genotype probabilities input in the AlphaGenes format (``-penetrance``). 
+The ``-penetrance`` is assumed to contain error-free phased? genotype probabilities. 
+Hence, when using ``-penetrance`` as the input, you should not use with options ``-est_geno_error_prob``,
```
gregorgorjanc: @XingerTang I am rereading these changes - so, we input phased geno probs, which gives us an initial state. I could as well input observed genotype data and when we propagate that observed information the phased geno probs will change from the initial state. I am not looking at the code, just trying to understand the logic.

If the above is correct, why shouldn't I estimate geno_error_prob or any other model parameters? In principle I could, but we should likely caution users that this might be unwise so they avoid double counting the information (if phased geno probs came from the same data) - I assume this is what you wanted to warn here!? Can you rephrase to this effect or lets clarify what the real "issue" for this "warning note" is.

src/tinypeel/tinypeel.py
```diff
@@ -751,8 +742,8 @@ def getArgs():
         required=False,
         type=str,
         nargs="*",
-        help=argparse.SUPPRESS,  # This argument will not appear in the help message, but is still available to use.
-    )  # help='An optional external penetrance file. This will overwrite the default penetrance values.')
+        help="Optional external penetrance file(s) (see format details in the docs). This will overwrite the default penetrance values.",
```
gregorgorjanc: Sync

gregorgorjanc:
@XingerTang I have now evaluated the used options and how we currently document inputs/outputs and agree with your summary - I do think we should likely want to have one file format explained once (say phased geno probs) and then we say how this is used differently on input and on output, but that is an issue for another day. Will open that.

Otherwise, we are close here. I suggested some edits and clarifications.

docs/source/usage.rst
XingerTang: I didn't write the warnings myself, it is there from previous developers. 
I guess the reason why we can't do the estimation with input phased geno prob is that the errors are estimated by the difference between genotype probabilities and genotypes; we are not guaranteed to have genotypes input if we are using genotype probabilities as the input. 
Even if we have the genotype input, the genotype error is used in a way that updates/rewrites the genotype probabilities based on the genotypes and updated genotype errors per peeling cycle. The penetrance input would be fully overwritten after a single cycle.

docs/source/usage.rst
XingerTang: The ``hap_file`` is read into the pedigree by ``tinyhouse.Pedigree.Pedigree.readInPhase``, the information would be stored in the ``haplotypes`` attribute of the ``Pedigree`` object before the peeling. 

Later, at line 88 of ``PeelingInfo.py``, 
```
HaplotypeOperations.ind_fillInGenotypesFromPhase(ind)
```
It will convert the haplotypes to genotypes (ignoring phasing information) and calculate genotype probabilities based solely on the genotypes.

docs/source/usage.rst
gregorgorjanc: > I didn't write the warnings myself, it is there from previous developers.
Noted.

> I guess the reason why we can't do the estimation with input phased geno prob is that the errors are estimated by the difference between genotype probabilities and genotypes; we are not guaranteed to have genotypes input if we are using genotype probabilities as the input.
Hmm, but the same applies when we use some observed genotype data just for some individuals, so I think this is not a blocker, in principle. I guess the question is what happens when we don't provide genotype data at all!? Can you check what happens in that case (with -est_geno_error...) - I assume that should fail or produce strange result!?

> Even if we have the genotype input, the genotype error is used in a way that updates/rewrites the genotype probabilities based on the genotypes and updated genotype errors per peeling cycle. The penetrance input would be fully overwritten after a single cycle.
Yes, this is what I had thought - we usually read in genotypes and calculate genotype probs based on that and pedigree. Here with genotype probs as input, we have the genotype probs immediately, however if we would bring in some additional genotype data (say the new batch), then by propagating that new data up the pedigree and then the updated genotype probs down the pedigree, we would get new/updated genotype probabilities. So, this makes sense, in principle, right?

I do agree we need some sort of a warning, the question just is how to word it. Maybe we write in 2-4 sentences what this import does and what happens so people will understand the implications! I will work with you on this, just draft something and we can iterate.

docs/source/usage.rst
gregorgorjanc: Ah, ok, so this is then not very powerful. OK, I propose the following.

1) Write somewhere what happens with these haplotypes - that they are imported but eventually combined and used as genotypes. Not powerful (since phase information is lost), but a capability if anyone wants it.

2) Open an issue to develop a function that calculates phased genotype probabilities from the haplotypes - this should not be hard and will not throw away the phase information. Do you agree?

docs/source/usage.rst
XingerTang: AlphaPeel would automatically set the parameter estimation flags to false if the phased genotype probability is provided, with an error message. 

In ``PeelingInfo.py``:
```
    if args.phased_geno_prob_file is not None:
        if peelingInfo.isXChr:
            warnings.warn(
                "Using an external phased genotype probability file and the x_chr option is highly discouraged. Please do not use."
            )

        if args.est_geno_error_prob:
            warnings.warn(
                "External phased genotype probability file included, but est_geno_error_prob flag used. The two options are incompatible. est_geno_error_prob set to false."
            )
            args.est_geno_error_prob = False

        if args.est_seq_error_prob:
            warnings.warn(
                "External phased genotype probability file included, but est_seq_error_prob flag used. The two options are incompatible. est_seq_error_prob set to false."
            )
            args.est_seq_error_prob = False
```

docs/source/usage.rst
```diff
@@ -87,12 +94,25 @@ one or more genomic data files to run the analysis.
...
+|Software| supports genotype probabilities input in the AlphaGenes format (``-phased_geno_prob_file``). 
+The ``-phased_geno_prob_file`` is assumed to contain error-free phased genotype probabilities. 
+Hence, when using ``-phased_geno_prob_file`` as the input, you should not use with options ``-est_geno_error_prob``, 
+``-est_seq_error_prob``. 
+The option does not support the ``-x_chr`` option for now. 
+
 When ``-x_chr`` is used the :ref:`pedigree file <ped_file_format>` and
 :ref:`genotype file <geno_file_format>` have specific requirements.
-Follow the links to respective file formats for more details.
+Follow the links to e file formats for more details.
```
gregorgorjanc:
```suggestion
Follow the links to file formats for more details.
```

docs/source/usage.rst
```diff
@@ -524,6 +544,36 @@ Example with four individuals and their allele read counts at four loci:
   id4 2 0 6 7
   id4 0 7 7 0
 
+.. _hap_file_format:
+
+Haplotype file
+==============
+
+This file has two lines with
+*observed phased haplotypes* for each phased genotyped individual;
```
gregorgorjanc:
```suggestion
*observed haplotypes* for each phased genotyped individual; 
```

docs/source/usage.rst
```diff
@@ -524,6 +544,36 @@ Example with four individuals and their allele read counts at four loci:
...
 
+.. _hap_file_format:
+
+Haplotype file
+==============
+
+This file has two lines with
+*observed phased haplotypes* for each phased genotyped individual; 
+possibly from previous analyses.
+The sequence of the maternal and paternal haplotypes won't affect the peeling.
```
gregorgorjanc:
```suggestion
The order of maternal and paternal haplotypes is not important
because the current implementation translates these haplotypes into genotypes.
```

docs/source/usage.rst
```diff
@@ -35,13 +35,20 @@ Input options
...
       -pheno_file [PHENO_FILE ...]
                           Phenotype file(s)
                           (:ref:`see format details <pheno_file_format>`).
       -plink_file [PLINK_FILE ...]
                           Plink (binary) file(s)
                           (:ref:`see format details <plink_file_format>`).
+      -phased_geno_prob_file [PHASED_GENO_PROB_FILE ...]
+                          Optional external phased_geno_prob_file file(s) 
+                          (:ref:`see format details <penetrance_file_format>`).
+                          This will overwrite the default genotype probability values.
```
gregorgorjanc:
```suggestion
                          This will provide the starting internal genotype probability state. 
```

Would this work better in line with our discussion?

src/tinypeel/tinypeel.py
```diff
@@ -627,6 +618,14 @@ def getArgs():
         type=str,
         help="Map file for loci in genomic data files (see format details in the docs).",
     )
+    input_parser.add_argument(
+        "-phased_geno_prob_file",
+        default=None,
+        required=False,
+        type=str,
+        nargs="*",
+        help="Optional external phased genotype probability file(s) (see format details in the docs). This will overwrite the default genotype probability values.",
```
gregorgorjanc:
```suggestion
        help="Optional external phased genotype probability file(s) (see format details in the docs). This will overwrite the starting internal genotype probability state.",
```

gregorgorjanc (changes requested):
Minor bits and we merge this one in @XingerTang!

docs/source/usage.rst
gregorgorjanc: Ok, I see now. Let’s keep the behaviour and adapt it if and as needed in the future.

gregorgorjanc (approved):
Looks good @XingerTang! Yay, this will close a bunch of issues;) Squash and merge.

